Issue #57, opened 2013-12-18
Memory leak


noplay:
```
About my case, I need to read new datas from a few tables of an mysql slave server. I am using you library only on WriteRowEvent, catching datas and inserting them in a cache db (Redis).
The program runs always on a server, waiting the next filtering event.

Standard libraries in the program are sys, os, logging, OptionParser, cPickle etc..
At this point of my development, I do not begin the Redis part job.
For debugging, I am using  WMI lib (for Win) to see memory size at each new event
If you need, I can send you the program and logs.

Yves

I am using the mysqlreplication lib to fill a cache memory (Redis) from 
a few tables of a slave mysql server. 
Program is always running, waiting for WriteRowEvent event about these 
tables. At this point of the developement, I do not write the Redis part. 
Program is using std lib as logging (TimedRotatingFileHandler), 
OptionsParser (command line args), CPickle and WMI (to trace memory size 
wih WinXYZ). Last source code and log file could be availables 

Yves 
```

romuald:
Shouldn't this issue be fixed ?

toontong:
issue do not fixed yet.  code did not merged? why?

noplay:
Code has been merged #98

toontong:
But, I use the master branch, MySQL.5.7 + python2.7 , 4000+tables, memory leek,

noplay:
Can you share your code?

Le ven. 29 juil. 2016 à 03:12, Huang ChuanTong <notifications@github.com> a
écrit :

toontong:
memory leak code was here: https://github.com/toontong/py-mysql-slaver/blob/master/pyslaver.py

when I merge your branch:memory leak to [my branch](https://github.com/toontong/python-mysql-replication/commit/f2628bfc9c4c72cf0bf2e5e9500cb4d07aaceab4).   no memory leak yet.

PR #59, opened 2014-01-05
Fix memory issues with large number of tables

julien-duponchelle:
This the pull request is about memory leak reported by @YvesDup in #57.

I think this need to be merge in binlogstream before merge of the pull request.


julien-duponchelle:
Too old I 'm not going to rebase it

PR #78, opened 2014-07-14
Fix 'utf8mb4' decode error,support 'utf8mb4' charset.

wmttom:
Fix decode error,if column.character_set_name == "utf8mb4".
Connect MySQL use 'utf8mb4',when maxuncode is 1114111(UCS4).


wahab013:
Hi 
I am facing the same issue and I have tried this pull request solution and that was working for me 
can you please guide me why you are not merging this PR

PR #82, opened 2014-08-06
Failing test to demo bug

darioush:
(Broken test, not for merging)

The code in `row_event.py` : `TableMapEvent::__init__` depends on being able to obtain the column schema from the active connection to the database. This fails if the table is deleted before the TableMap event is processed.

Potential fixes include (a) reading column schemas from create / alter tables, (b) passing in a null / null-like schema for deleted tables.


bjoernhaeuser:
@darioush do you want to implement this?

PR #91, opened 2014-09-27
Use a simplifed  byte2int

julien-duponchelle:
It's offer a very small performance gain

The original implementation from PyMySQL test instance type
before.
https://github.com/PyMySQL/PyMySQL/blob/3be54819ee6b09710be892a2b86f4592eaabf412/pymysql/util.py

It's not require and by dropping it's a little faster.

### DIFF
--- a/pymysqlreplication/event.py
+++ b/pymysqlreplication/event.py
@@ -3,7 +3,7 @@
 import struct
 import datetime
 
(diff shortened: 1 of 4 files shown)


bjoernhaeuser:
Hm, I dont like it.

PR #101, opened 2014-12-05
expose binlog event signal

lxyu:
this will allow blinker.signal hook on binlog events and allow a better code customization


bjoernhaeuser:
Gonna close this, the PR came to a better solution through discussion.

PR #105, opened 2015-03-27
Add query property

cccc1999:
I am trying to do unit test with 
mock_query_event = mock.Mock(spec=QueryEvent)
it would not let me use mock_query_event.query, so I am suggesting that we turn it into a property!

Thanks!


bjoernhaeuser:
Can you check why travis fails the test? And could you also please squash your commits? We dont want to have 3 Commits for adding 4 lines :-)

PR #112, opened 2015-05-21
add support for BeginLoadQueryEvent and ExecuteLoadQueryEvent, update tests

jasonzzz:


noplay:
Cool,

Do you think it's possible to write a test in order to avoid breaking it in another release?

jasonzzz:
Hi, this case happens only when `binlog-format = statment`, but I find this [line](https://github.com/noplay/python-mysql-replication/blob/master/.travis.yml#L36) set `binlog-format = row`.

How can I change mysql config when testing this case using Travis CI?

noplay:
You are right changing the config during the test is hard :(

noplay:
I don't think we have an esay solution. Perhaps with Travis build matrix but it's start to be complicated. You can comment the test.

jasonzzz:
I've commented the test. 

It builds [success](https://travis-ci.org/jasonzzz/python-mysql-replication/builds/63622067) under my repository though one [failed](https://travis-ci.org/noplay/python-mysql-replication/builds/63622085) here.

noplay:
Thanks !

PR #114, opened 2015-05-29
Support for skipping the binlog until reaching specified timestamp.

vartec:
The rationale behind this is that as server fails over, the binlog position you've got stored from failed server might no longer be valid on the new one. With MySQL 5.6 you're fine, because you got GTIDs, but if you're using older MySQL there is no way of telling from which log position you should start. Timestamp  helps to _roughly_ know where to start. If your app is capable of dealing with duplicates, that is good enough.

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -40,7 +40,7 @@ class BinLogStreamReader(object):
                  filter_non_implemented_events=True,
                  ignored_events=None, auto_position=None,
                  only_tables=None, only_schemas=None,
-                 freeze_schema=False):
+                 freeze_schema=False, skip_to_timestamp=None):
         """
         Attributes:
             resume_stream: Start for event from position or the latest event of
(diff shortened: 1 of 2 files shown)


noplay:
Thanks! I'm away from home I will take a look on that when coming back at
the beginning of next week.

Le ven. 29 mai 2015 03:48, Bartek Ogryczak <notifications@github.com> a
écrit :

bjoernhaeuser:
Why not implement that just in the application itself?

noplay:
I think it's usefull if you changed the schema and want to skip the events.

vartec:
@bjoernhaeuser the reason for doing here and not at application level is to avoid inflating rows into Python objects, which is CPU consuming and seriously limits the throughput. 

MySQL's binlog reader has similar option: https://dev.mysql.com/doc/refman/5.6/en/mysqlbinlog.html#option_mysqlbinlog_start-datetime

vartec:
sorry about this back and forth with tests, but the outcome seems to depend heavily on MySQL version and configuration :-/

vartec:
FYI, I've squashed the unittest commit mess.

noplay:
Thanks

PR #115, opened 2015-05-29
skipping GTID tests on DBs not set up to support GTID

vartec:
It's been painful debugging unittests on DB which doesn't support GTID either because it's MySQL older than 5.6 or because it `gtid_mode` is set to OFF. This should help.


bjoernhaeuser:
I like this :) @noplay do we want to merge this?

noplay:
Nice :)

PR #117, opened 2015-06-17
Partial fix for dropped columns blowing up replication 

vartec:
Partial fix for dropped columns blowing up replication when replaying binlog with past events

Mitigates following error:
```
  File "/usr/local/lib/python2.7/dist-packages/pymysqlreplication/binlogstream.py", line 262, in fetchone
    self.__freeze_schema)
  File "/usr/local/lib/python2.7/dist-packages/pymysqlreplication/packet.py", line 98, in __init__
    freeze_schema = freeze_schema)
  File "/usr/local/lib/python2.7/dist-packages/pymysqlreplication/row_event.py", line 550, in __init__
    column_schema = self.column_schemas[i]
IndexError: list index out of range
```

This is just a superficial patch. It does not address deeper issue at hand, which is that `pymysqlreplication` has no good way of handling `RowEvent`s containing columns which have been dropped prior to `pymsqlreplication` start. The problem is always relying on current version of `information_schema` even though dealing with events that might be several versions in the past.

### DIFF
--- a/pymysqlreplication/row_event.py
+++ b/pymysqlreplication/row_event.py
@@ -547,7 +547,20 @@ class TableMapEvent(BinLogEvent):
             self.packet.read_length_coded_binary()
             for i in range(0, len(column_types)):
                 column_type = column_types[i]
-                column_schema = self.column_schemas[i]
+                try:
+                    column_schema = self.column_schemas[i]
+                except IndexError:
+                    # this a dirty hack to prevent row evens containing columns which have been dropped prior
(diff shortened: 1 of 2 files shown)


vartec:
I've also created issue this and the underlying problem: https://github.com/noplay/python-mysql-replication/issues/118

noplay:
Do you think it's possible to write a test for this ? I'm not sure

vartec:
@noplay I've added unittest. Verified that it fails on your master.

noplay:
Great thanks !

PR #120, opened 2015-06-27
Column attributes as regular attributes instead of a dict

vartec:
Making Column attributes into regular object attributes instead of a dictionary.

This makes getting the attribute about 20 times faster.

Rationale behind this is the fact that 15% of time spent on creating and using
`RowsEvent` was spent on `Column.__getattr__`. 

There is more to be optimized, but this one was a low hanging fruit.

### DIFF
--- a/pymysqlreplication/column.py
+++ b/pymysqlreplication/column.py
@@ -10,76 +10,75 @@ class Column(object):
     """
 
     def __init__(self, *args, **kwargs):
-        self.data = {}
         if len(args) == 3:
             self.__parse_column_definition(*args)
         else:
-            self.data = kwargs
+            self.__dict__.update(kwargs)
 
     def __parse_column_definition(self, column_type, column_schema, packet):
-        self.data["type"] = column_type
-        self.data["name"] = column_schema["COLUMN_NAME"]
-        self.data["collation_name"] = column_schema["COLLATION_NAME"]
-        self.data["character_set_name"] = column_schema["CHARACTER_SET_NAME"]
-        self.data["comment"] = column_schema["COLUMN_COMMENT"]
-        self.data["unsigned"] = False
-        self.data["type_is_bool"] = False
+        self.type = column_type
+        self.name = column_schema["COLUMN_NAME"]
+        self.collation_name = column_schema["COLLATION_NAME"]
+        self.character_set_name = column_schema["CHARACTER_SET_NAME"]
+        self.comment = column_schema["COLUMN_COMMENT"]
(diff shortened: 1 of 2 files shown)


bjoernhaeuser:
Do you still have your benchmarks around? :)

vartec:
@bjoernhaeuser I'll try to make isolated benchmark. What I have is cProfile of my application which led me to make that change:
```
         5347610 function calls (5343431 primitive calls) in 31.533 seconds

   Ordered by: internal time
   List reduced from 1560 to 78 due to restriction <0.05>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  124/123    6.366    0.051    6.412    0.052 {method 'recv_into' of '_socket.socket' objects}
     3600    4.946    0.001   19.533    0.005 /usr/local/lib/python2.7/dist-packages/pymysqlreplication/row_event.py:66(_read_column_data)
  1681477    2.926    0.000    2.926    0.000 /usr/local/lib/python2.7/dist-packages/pymysqlreplication/column.py:94(__getattr__)
     7294    2.114    0.000    2.114    0.000 {method 'acquire' of 'thread.lock' objects}
   242165    1.560    0.000    3.412    0.000 /usr/local/lib/python2.7/dist-packages/pymysqlreplication/packet.py:102(read)
    39600    1.228    0.000    6.352    0.000 /usr/local/lib/python2.7/dist-packages/pymysqlreplication/row_event.py:317(__read_new_decimal)
```

And simple experiment:

```
import timeit


class ObjectWithDict(object):
    def __init__(self):
        self.data = {"attr": "value"}

    def __getattr__(self, item):
        try:
            return self.data[item]
        except KeyError:
            raise AttributeError("{0} not found".format(item))


class ObjectWithAttr(object):
    def __init__(self):
        self.attr = "value"


if __name__ == "__main__":
    t_dict = timeit.timeit("_ = obj.attr", "import __main__; obj = __main__.ObjectWithDict()")
    t_attr = timeit.timeit("_ = obj.attr", "import __main__; obj = __main__.ObjectWithAttr()")

    print "t_dict {}, t_attr {}, speedup {}".format(t_dict, t_attr, t_dict/t_attr)
```

Which with CPython 2.7 yields result:
`t_dict 0.752788066864, t_attr 0.036523103714, speedup 20.6112841`

bjoernhaeuser:
@vartec uhm, huge difference in there already :) I am okay with merging. What do you think @noplay? :)

noplay:
:D

PR #122, opened 2015-07-03
gngtid failed parsing: raise with value

romuald:
When a gtid value is incorrect raise the value that we attempted to parse


noplay:
Thanks

PR #124, opened 2015-07-03
Explicit close stream connection on exception

giacomolozito:
Issue https://github.com/noplay/python-mysql-replication/issues/121 for details.
This change helps with PyPy, which does not seem to close the socket automatically on exception, leaving a socket in CLOSE_WAIT for each disconnection until the program is terminated.


noplay:
Thanks !

PR #126, opened 2015-07-08
Fix GtidSet __str__ representation

romuald:
Gtid str representation is built from the str of their gtid, and not their repr


noplay:
Thanks !

PR #129, opened 2015-08-13
Don't fail on incomplete dates (with 00 as day or month)

hc:
Avoid total failure when encountering dates such as 2015-00-21 2015-05-00


noplay:
Thanks a lot

Sorry for the late reply I was in vacations

PR #134, opened 2015-12-04
Close connection on EOF packet

romuald:
Since client might want to reconnect, close connection and reset connection status when receiving an EOF packet

(note current code seems to work correctly without this patch :p)


noplay:
Cool thanks

PR #135, opened 2016-01-01
bugfix for decoding values of columns in charset "utf8mb4"

jffifa:
Python would throw an exception for decoding values of columns in charset "utf8mb4", as there is no such encoding "utf8mb4" in python, but just "utf8".


noplay:
Thanks!

PR #136, opened 2016-01-09
To fix the bug caused by  DictCursor mode.  

dikoufu:
DictCursor in MYSQL_SETTINGS  ,will appear an Error: 
TypeError: unhashable type.

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -113,7 +113,10 @@ class BinLogStreamReader(object):
 
         if result is None:
             return False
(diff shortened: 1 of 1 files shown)


dikoufu:
To fix the bug caused by  DictCursor mode.  When cursorclass is DictCursor in MYSQL_SETTINGS  ,will appear an Error: TypeError: unhashable type.

bjoernhaeuser:
Uh, this looks more like a workaround then a fix. Where is this coming from?

dikoufu:
No mysql connect ,so check failed?

Issue #132, opened 2015-09-16
Decoding binary as utf-8


MartinAmps:
I get this trace:

Traceback (most recent call last):
  File "./test.py", line 33, in <module>
    main()
  File "./test.py", line 26, in main
    for binlogevent in stream:
  File "/usr/local/lib/python2.7/dist-packages/pymysqlreplication/binlogstream.py", line 262, in fetchone
    self.__freeze_schema)
  File "/usr/local/lib/python2.7/dist-packages/pymysqlreplication/packet.py", line 98, in __init__
    freeze_schema = freeze_schema)
  File "/usr/local/lib/python2.7/dist-packages/pymysqlreplication/event.py", line 141, in __init__
    self.query = tmp.decode("utf-8")
  File "/usr/lib/python2.7/encodings/utf_8.py", line 16, in decode
    return codecs.utf_8_decode(input, errors, True)
UnicodeDecodeError: 'utf8' codec can't decode byte 0xae in position 460: invalid start byte

I printed a repr of the packet and it is essentially:

    INSERT INTO x(......, ip) VALUES (....,  'b\xae\xe1\xbd');

This is row based replication where the master was originally sent INET6_ATON('::1') for example.

What's the recommended solution here? I'm surprised no one else has hit this as many column types leverage binary.

Thanks!

noplay:
Can you post the type of the column?

Le jeu. 17 sept. 2015 à 00:56, Martin Amps <notifications@github.com> a
écrit :

MartinAmps:
Apologies - it’s a varbinary(15)

noplay:
No problem :) Thanks a lot for the report. Sorry if I was rude in my
question I just wake up :P

Le jeu. 17 sept. 2015 à 08:15, Martin Amps <notifications@github.com> a
écrit :

> Apologies - it’s a varbinary(15)
>
> > On Sep 16, 2015, at 11:13 PM, Julien Duponchelle <

MartinAmps:
No problem at all. Let me know if I can help at all debugging further - right now I just wrapped it in a try: except: block to make it fail a bit more gracefully, was planning to dive in a little deeper tomorrow. Bed time here on the west coast!

noplay:
I think we can ignore the invalid unicode char (do not break existing app) or return a byte string

MartinAmps:
I agree, it would probably be useful to parse it out in future but for now the potential exceptions should be fixed

PR #146, opened 2016-03-11
added fix for non utf8 query event

mastak:
I faced with this #132 problem. 

Does this solution have a chance to live? (=


noplay:
What is the schema of the column creating the issue?

bjoernhaeuser:
@mastak can you add a test which show cases the error (and the fix)?

mastak:
Sorry, guys, that I'm so slowly.
The column is simple varchar, the table has latin-1 charset. 
I wrote some test, but it is work only with STATEMENT based replication, and I don't know how it can be enabled. Can you help with it?

```python
    def test_encoding_latin1_statement(self):
        db = copy.copy(self.database)
        db["charset"] = "latin1"
        self.connect_conn_control(db)

        create_query = "CREATE TABLE test (test CHAR(12)) CHARACTER SET latin1 COLLATE latin1_bin;"
        insert_query = b"INSERT INTO test VALUES('\x96')"

        self.execute(create_query)
        self.execute(insert_query)
        self.execute("COMMIT")

        self.assertIsInstance(self.stream.fetchone(), RotateEvent)
        self.assertIsInstance(self.stream.fetchone(), FormatDescriptionEvent)
        # QueryEvent for the Create Table
        self.assertIsInstance(self.stream.fetchone(), QueryEvent)

        # QueryEvent for the BEGIN
        self.assertIsInstance(self.stream.fetchone(), QueryEvent)

        event = self.stream.fetchone()
        self.assertIsInstance(event, QueryEvent)

        self.assertEqual(event.query, insert_query.decode("latin-1"))
```

mastak:
Could we do something like:
at the begin of the test `SET GLOBAL binlog_format = 'STATEMENT';`
at the end of the test  `SET GLOBAL binlog_format = 'ROW';`
?

noplay:
Clearly the fix work but will break it's non utf-8 and non latin1. 

But what do you think about @baloo solution in #159 to just return a byte object so you can handle it in your code?

mastak:
Sure, it would be good.

mastak:
We can't return bytes (even only in case with error) because it will break existing systems.

But what if add some method like:
```python
def _decode_query(self, query)
    return query.decode("utf-8")
```

and then everyone can overwrite it. What  do you think?

PR #147, opened 2016-03-11
added intvar event

mastak:
Travis contain row level settings for replication, so it cannot be tested, is not it?


noplay:
Thanks, yes it's hard to test

PR #148, opened 2016-03-14
gtid: fixup parsing on mysql format

baloo:
Mysql may return gtid joined by ',\n' separator:
 SELECT @@global.gtid_executed


noplay:
Thanks

PR #151, opened 2016-03-30
Enabling IntvarEvents and master server validation

zanaca:
- Enable parsing intvar events
- Verifies that the informed mysql server is properly configured as master


noplay:
The code is good  but the build has broke. I think it's trivial to fix.
https://travis-ci.org/noplay/python-mysql-replication/jobs/119599561

Thanks a lot for the help

bjoernhaeuser:
@zanaca Any chance you update your PR and fix the tests?

PR #153, opened 2016-04-12
setup travis to run test on multiple mysql version (5.6 and 5.7)

26fe:
With this patch tests can run on different mysql version (5.6 and 5.7). 
Some note about the patch:
- scripts directory store install script for different mysql version
- moved install instruction from .travis.yml to scripts/install_mysql56.sh
- scripts/install_mysql_sandbox.sh install [mysql_sandbox](http://mysqlsandbox.net/)
- script/install_mysql_57_on_sandbox.sh use mysql-sandbox to install mysql 5.7.12
- the env part define DB enviroment variable that decide on which mysql version test will be run

Travis run every combination from mysql version and python version.

The mysql_sandbox could be used to install other version (maybe 5.5). Using the sandbox is more generic that using the s.o. package manager and we can choice the exact version of mysql instead of the provided. version of the package maintainer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,56 +5,16 @@ python:
   - "3.3"
   - "3.4"
   - "pypy"
+env:
+  - DB=mysql57
+  - DB=mysql56
 install:
-    - "pip install ."
-    - "pip install nose"
+  - "pip install ."
+  - "pip install nose"
 cache: apt
 before_script:
-  # Remove old mysql version
-  - "sudo /etc/init.d/mysql stop || true"   
-  - "sudo apt-get remove mysql-common mysql-server-5.5 mysql-server-core-5.5 mysql-client-5.5 mysql-client-core-5.5"
-  - "sudo apt-get autoremove"
-  # Config
-  - "sudo sed -i'' 's/table_cache/table_open_cache/' /etc/mysql/my.cnf"
-  - "sudo sed -i'' 's/log_slow_queries/slow_query_log/' /etc/mysql/my.cnf"
-
-  # Install new mysql version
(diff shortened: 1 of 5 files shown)


bjoernhaeuser:
That is awesome <3

bjoernhaeuser:
What is this mysql-sandbox-stuff?

noplay:
Seem to be a perl project making magic :) http://search.cpan.org/~gmax/MySQL-Sandbox-3.1.00/lib/MySQL/Sandbox.pm#Single_server_sandbox

26fe:
Mysql-sandbox is a well established utility to setup MySQL instances for test. It takes care of the little differences between mysql versions.  If I want test python-mysql-replication on travis with more mysql, the alternatives, that I am able to think, are:

1. installing mysql version 5.7 using the package manager. I have tried and it is a headache. But the major disadvantage is you are limited by the mysql version used by package maintainer. And anyway it is not easy to change linux distribution. 
2. installing mysql version 5.7 directly. It is no so difficult, but the script must be reviewed when a different version is needed.
3. installing mysql using mysql-sandbox. It adds a bit of complexity (other package must be downloaded). But we have a consistent configuration between mysql version on the same machine and on different machines.  I use the same setup on my laptop.
I prefer the 3 :-)

The url fo mysql sandbox is http://mysqlsandbox.net/. The blog of author is http://datacharmer.blogspot.it/.

26fe:
I was writing while you have merged. Thank you for accepting my pull request!

noplay:
Thanks a lot it's a precious contribution

Le mer. 13 avr. 2016 à 08:55, Giovanni F. <notifications@github.com> a
écrit :

PR #155, opened 2016-05-17
support read binlog for both mysql and mariadb

dna2github:
- change command from 0x1e to 0x12
- add two parameters: binlog_name and binlog_pos to read binlog
  from specified position


dna2github:
the binlog_name and binlog_pos do not work, pls ignore them... :(

baloo:
This would break mysql >= 5.6 when using `auto_position`.

baloo:
I'm pretty sure what you're trying to achieve is already implemented in `log_file` and `log_pos`

PR #158, opened 2016-05-20
unittest2 is only a test requirement

asottile:


noplay:
Thanks,

It seem to failed on travis for Python 2.6 and 2.5

bjoernhaeuser:
Suggesting to remove python 2.6 support. Its actually not maintained anymore.

asottile:
I guess there's three options:

1. Change `nosetests` to `python setup.py test` (this'll honor `tests_require` but use setuptools installation (kinda icky)
2. Add an `extras_require` for `test` and use `pip install .[test]` and use the equivalent workflow
3. Drop python2.6 (removing unittest2 mentions entirely, and from .travis.yml)

Please advise :D

noplay:
If I remember some user use Python 2.6 for Red Hat 6.

But I'm not against stop testing on 2.6 and mark it as not supported.

Le ven. 20 mai 2016 à 16:49, Anthony Sottile <notifications@github.com> a
écrit :

asottile:
I opted for option 2, let me know what you guys think of the new diff :)

baloo:
LGTM

bjoernhaeuser:
👍

PR #160, opened 2016-07-02
Fix the example: Add required server_id

dveeden:
Otherwise this happens:
```
Traceback (most recent call last):
  File "./test.py", line 6, in <module>
    stream = BinLogStreamReader(connection_settings = mysql_settings)
TypeError: __init__() missing 1 required positional argument: 'server_id'
```


bjoernhaeuser:
Thanks!

PR #161, opened 2016-07-13
fixed count of fetched events

mastak:


noplay:
Hi,

Why do you need to increase the value?

mastak:
Hi,

I think ranges in line 109 and line 122 must be the same.

noplay:
Gotcha!

PR #162, opened 2016-08-01
use 0 as date/datetime zero value

xiezhenye:
return 0 as zero value of date/datetime to tell difference from NULL and '0000-00-00'.
e.g. insert/update a NOT NULL date column, only '0000-00-00' or 0 is valid.

### DIFF
--- a/pymysqlreplication/row_event.py
+++ b/pymysqlreplication/row_event.py
@@ -250,13 +250,13 @@ class RowsEvent(BinLogEvent):
     def __read_date(self):
         time = self.packet.read_uint24()
         if time == 0:  # nasty mysql 0000-00-00 dates
-            return None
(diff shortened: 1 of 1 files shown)


noplay:
That sound correct.

But I worry that will force people to handle this case in their code. What is the good usage of 0000-00-00?

What others think about it? 

PS: it seem some tests still failed

baloo:
Looks good to me, just need to make sure this is not a patch-level release. And it's shown in the changelog.

noplay:
@xiezhenye can you fix the tests? I will merge it after that

Thanks for the help

xiezhenye:
MySQL 5.6 tests passed. MySQL 5.7 tests seems failed to start test instance.

Issue #56, opened 2013-12-18
Strange error in event_row.py


bohnster:
Hi,

Nice library. 

Here is a problem that has started happening, and I don't know why. Here is the Python error:

 File "/usr/lib/python2.6/site-packages/pymysqlreplication/binlogstream.py", line 146, in fetchone
    self.**use_checksum)
  File "/usr/lib/python2.6/site-packages/pymysqlreplication/packet.py", line 93, in __init**
    ctl_connection)
  File "/usr/lib/python2.6/site-packages/pymysqlreplication/row_event.py", line 491, in **init**
    column_schema = self.column_schemas[i]
IndexError: tuple index out of range

I think this is due to some MySQL internal query that pymysqlreplication has trouble with. The EventType that appears to trigger the failure is a QueryEvent, but the event is an internal MySQL event. Here some events that occur just before the failure. You can see that these are internal MySQL queries:

=== QueryEvent ===
Date: 2013-12-11T16:31:15
Log position: 30510
Event size: 80
Read bytes: 80
Schema: mysql
Execution time: 0
Query: TRUNCATE TABLE time_zone_transition
=== QueryEvent ===
Date: 2013-12-11T16:31:15
Log position: 30614
Event size: 85
Read bytes: 85
Schema: mysql
Execution time: 0
Query: TRUNCATE TABLE time_zone_transition_type
=== QueryEvent ===
Date: 2013-12-11T16:31:15
Log position: 30683
Event size: 50
Read bytes: 50
Schema: mysql
Execution time: 0
Query: BEGIN

^^ error occurs immediately after this

You can see that the date is 2013.12.11.

I looked at the source code where the error occurs:

485        for i in range(0, len(column_types)):
486            column_type = column_types[i]
487            column_schema = self.column_schemas[i]
488            col = Column(byte2int(column_type), column_schema, from_packet)
489            self.columns.append(col)

The problem is that self.column_schemas is None, so index fails. 

Any ideas on this? Thanks.

dmariassy:
Hi,

and thanks for the very handy library! 👍

Although this thread has been closed for a while, I think that the issue pointed out by @banks continues to pose a serious problem for many potential users. The ability to reliably replay events that were created before the streaming process started (or while the process was temporarily down) is required in almost all real-world use-cases.

But I think that adding this functionality would not be that difficult if we (as suggested by banks and implemented by [debezium.io](http://debezium.io/blog/2016/04/15/parsing-ddl/)):

[...]

PR #163, opened 2016-08-07
Decoupling table mapping to allow replaying all binlog events

dmariassy:
As discussed in #56, starting a stream at a point in time where the schema of the source database was different than what it is now, will always break the program.

We can address this issue by allowing users of the library to set up a secondary MySQL database - a so-called Schema DB - and:
1. Catching all those QueryEvents that modify the underlying schema
2. Executing them on the Schema DB
3. Pointing the BinlogStreamReader's __get_table_information method to this DB's information schema.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -42,3 +42,9 @@ _build
 
 # Pyenv
 .python-version
+
+# Mac System files
+.DS_Store
+
+# Virtualenv folder
+env/
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -8,7 +8,7 @@ from pymysql.cursors import DictCursor
 from pymysql.util import int2byte
 
 from .packet import BinLogPacketWrapper
-from .constants.BINLOG import TABLE_MAP_EVENT, ROTATE_EVENT
+from .constants.BINLOG import TABLE_MAP_EVENT, ROTATE_EVENT, QUERY_EVENT
 from .gtid import GtidSet
 from .event import (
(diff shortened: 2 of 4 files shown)


noplay:
First it's very interesting.

I'm not really happy to add an external dependency like sqlparse because it's complicate licence stuff (I know at least one org where this lib is used and where they need to ask legal departement about dependencies ...).

Do you think we can found a way to move outside the lib this system? I think we can provide a way for user to replace __get_table_information by allowing him to pass a function  at init. Offering you the warranty that next version of pymysqlreplication will not break your code.

dmariassy:
Hi and thanks a lot for your feedback!

I understand your problem with external dependencies. However, simply allowing the user to pass in his/her own __get_table_information function wouldn't be sufficient to achieve the same result. That's because the parsing of the DDL statements must proceed at the same pace as the processing of other binlog events. Only this can ensure that the schema used for table mapping and the Row Events which are validated against the table map are always in sync. That's why I added the Query Event parsing and execution part to the main binlog streaming loop.

The solution I'm proposing consists of the three steps I listed in my original commit message, and the __get_table_information method only comes into play at the third stage of that process. I don't currently see how we could package all steps of the logic into a customisable __get_table_information function, since this function cannot be used to catch schema-changing Query Events.

So as I currently see it, if we really want to move away from sqlparse, our best bet would probably be an effort to mimic its capabilities and create our own (dumb) SQL parser that is only trained to catch DDL statements. But I'm not very keen on implementing that :)

WDYT?

baloo:
I believe it has been implemented and merged in #176. Feel free to reopen if I'm wrong.

PR #170, opened 2016-09-03
Fix build by updating to mysql 5.7.14

bjoernhaeuser:


bjoernhaeuser:
@baloo @noplay can you take a look? This updates mysql 5.7 to latest (which is available for download only), and makes the build green again :)

PR #174, opened 2016-09-27
Decode gtid from network packet

baloo:
I had to debug a weird replication behaviour. Wireshark was no help to read a COM_BINLOG_DUMP_GTID packet. It would probably help someone else :)


noplay:
Nice,

Do you think we can have a test for this?

baloo:
@noplay done

noplay:
Perfect

PR #176, opened 2016-11-04
Allow users to define ctl_connection_settings and the option to fail when table information is unavailable

### LINKED ISSUE #56: Strange error in event_row.py

kdparker:
Resolves #56 
To allow users to create their own secondary database that tracks schema changes if they want to ensure they will always have table_metadata (in case of a dropped table before a message can be picked up). It also adds an (optional) parameter that makes the BinLogStreamReader raise an exception if there is an issue getting schema information for a query. This allows the library to stay low-level while giving users the option of alleviating this issue.


kdparker:
By the way, if you want me to squash these commits, let me know.

.gitignore
```diff
@@ -37,6 +37,9 @@ _build
 *.xml
 *.iml
 
+# Temp vim .swp s
+*.swp
```
baloo: Pro tip: this should go in your global ignore file instead. Defaults to: $HOME/.config/git/ignore

This is not blocking for merge though.

pymysqlreplication/binlogstream.py
```diff
@@ -150,20 +152,28 @@ def __init__(self, connection_settings, server_id, resume_stream=False,
...
             report_slave: Report slave in SHOW SLAVE HOSTS.
             slave_uuid: Report slave_uuid in SHOW SLAVE HOSTS.
+            fail_on_table_metadata_unavailable: Should raise exception if we can't get
+                                                table information on row_events
         """
+
         self.__connection_settings = connection_settings
-        self.__connection_settings["charset"] = "utf8"
+        if not connection_settings.get("charset"):
+            self.__connection_settings["charset"] = "utf8"
```
baloo:
``` python
self.__connection_settings.setdefault("charset", "utf8")
```

noplay:
I agree the change sound good. If you can add a test for ctl_connection_settings it will be perfect

kdparker:
Tests are done and working now (sorry for commit spam earlier, they're all squashed now)

baloo:
\o/ I started looking at implementing the tests myself, and I have to admit I became quite lazy when looking at the travis mess. Thank you very much for your work. All looks good to me.

Issue #183, opened 2016-12-15
caution: misleading output if user do not have SELECT privilege on information_schema


danfengcao:
@imzcy1987 and I took a lot of time to uncover values missing if user do not have SELECT privilege on information_schema. Same issues have reported in #116 , #131 and #172 .

The root cause i found is that information_schema is a special db, Each MySQL user has the right to access these tables, but can see only the rows in the tables that correspond to objects for which the user has the proper access privileges([http://dev.mysql.com/doc/refman/5.7/en/information-schema.html](http://dev.mysql.com/doc/refman/5.7/en/information-schema.html) ). So in function [binlogstream.__get_table_information()](https://github.com/noplay/python-mysql-replication/blob/master/pymysqlreplication/binlogstream.py#L480), it always return empty either the table actually not existing or the user have no privilege on information_schema.COLUMNs.

I strongly suggest @noplay to update documentation  to indicate we should have
`GRANT SELECT, REPLICATION SLAVE, SUPER ON *.* TO user`

I also suggest to add an assertion before [https://github.com/noplay/python-mysql-replication/blob/master/pymysqlreplication/binlogstream.py#L471](https://github.com/noplay/python-mysql-replication/blob/master/pymysqlreplication/binlogstream.py#L471) , to raise Exception stating that user has  no SELECT privilege on information_schema.

I think it's important to pay attention to fix this bug or update documentation. Then #116 , #131 and #172 can be closed

noplay:
HI,

Pull requests are welcome. You can edit the documentation here:
https://github.com/noplay/python-mysql-replication/tree/master/docs

I think the permission check need to be done just one time at connection to the DB.

Issue #172, opened 2016-09-09
replication from other slave doesnt work


walito:
Hello, Im trying to run replication from other slave (log_slave_updates) but I cant get EVENTs, these are empty.

There is a BUG? or the replication doesnt work from other slave?

Thanks in advance

=== UpdateRowsEvent ===
Date: 2016-09-09T13:47:46
Log position: 1690915
Event size: 935
Read bytes: 13
Table: XXXX.ZZZZZ
Affected columns: 10
Changed rows: 0
Affected columns: 10
Values:
()

=== WriteRowsEvent ===
Date: 2016-09-09T13:47:46
Log position: 1689821
Event size: 160
Read bytes: 11
Table: XXXX.ZZZZZ
Affected columns: 10
Changed rows: 0
Values:
()

facundo-guerrero-olx:
It's the binlog in Row? I always using this from a Slave.

walito:
yes, Im using format row

MASTER ---------> SLAVE1 (log_slave_updates ON) ----------> SLAVE2 (mysql work fine)
                                                                                \---------> python-mysql-replication (doesnt work)


Thanks in advance

bjoernhaeuser:
@walito is your user able to read the information schema tables? See here https://github.com/noplay/python-mysql-replication/pull/184

In general we are used the same topology you have without any issue. Closing this therefore. Please reopen if the issue is not resolved.

Issue #131, opened 2015-09-08
Values missing if database rights aren't granted


foxycode:
Took me lot of time to uncover that user without database rights (on with  REPLICATION SLAVE and REPLICATION CLIENT) can't see table structure from information_schema and it's why event values aren't dumped.

It is possible read data from strem even without that rights?

bjoernhaeuser:
Nope, where would you get that information from?

foxycode:
@bjoernhaeuser Here data about table structure are read: https://github.com/noplay/python-mysql-replication/blob/master/pymysqlreplication/binlogstream.py#L332

If I haven't rights to that database/table, query returns empty result and row events doesn't return column data. I think that column data should be present in stream, even if I don't have database right. After all, how mysql slave could replicate them if they weren't. But I didn't examined your source too much to know how to handle it.

bjoernhaeuser:
@foxycode they are not included in the stream :( the stream misses the column name and data type, it only carries the position of these columns.

foxycode:
@bjoernhaeuser I understand that column names and data types aren't in stream. But it is possible to parse at least raw data and return them as tupple for example? Is it possible to decode them properly without knowing data type?

And if not, could you please mention information about needed rights in README somewhere so someone else don't need to deal with this?

Thanks

noplay:
Yeah you can imagine an option turning off the mapping to column name (the information schema is only need for that).

muayyad-alsadi:
please update documentation to indicate we should have

```
CREATE USER 'repl'@'%' IDENTIFIED BY 'password';
GRANT REPLICATION SLAVE ON *.* TO 'repl'@'%';
GRANT REPLICATION CLIENT ON *.* TO 'repl'@'%';
GRANT SELECT ON *.* TO 'repl'@'%';
```

bjoernhaeuser:
Done! https://github.com/noplay/python-mysql-replication/pull/184

Issue #116, opened 2015-06-17
No SELECT privileges result in silent failure


vartec:
Steps to reproduce:
1. `GRANT REPLICATION SLAVE, REPLICATION CLIENT ON *.* TO 'a_user'@'%'`
2. `REVOKE SELECT ON`your_db`.* TO 'a_user'@'%'` (or just don't grant it).
3. use pymysqlreplication 

Expected result: 
Raising an Exception stating that user has no privilege to read table's column information

Actual result:
Receiving RowEvents with no statement rows associated with it.

Background:
`BinLogStreamReader. __get_table_information` tries to access column information. Having no privileges this does not return error, but rather no rows. This information is later used in `RowEvent.__init__`. 
https://github.com/vartec/python-mysql-replication/blob/a4aa08c746cf69ed141301db81b6660655eff3c4/pymysqlreplication/row_event.py#L55-L58

 Having no column information causes it to silently marked as `completed = False`, and subsequently not fetching any rows.

bjoernhaeuser:
See also here: https://github.com/noplay/python-mysql-replication/pull/184

It's a design decision in the end, which resulted in user depending on the current behaviour. Nonetheless: patches welcome :)

PR #184, opened 2016-12-16
update doc to indicate privileges the user need

danfengcao:
As I comment in #183 , it's import to add privilege introduction to avoid the misleading output, which also be reported in #116 , #131 and #172 .


bjoernhaeuser:
Thank you very much for the contribution! I closed the referenced issues.

PR #188, opened 2017-01-17
Adding in JSON support for bin log output.

charltonaustin:
I added this because I thought it might be useful to have JSON as an output type for the bin logs. I don't know if you need or want it, but I thought I would offer it back.

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -7,6 +7,7 @@ from pymysql.constants.COMMAND import COM_BINLOG_DUMP, COM_REGISTER_SLAVE
 from pymysql.cursors import DictCursor
 from pymysql.util import int2byte
 
+from pymysqlreplication.representation import Human, string_to_types
 from .packet import BinLogPacketWrapper
 from .constants.BINLOG import TABLE_MAP_EVENT, ROTATE_EVENT
 from .gtid import GtidSet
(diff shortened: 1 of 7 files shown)


pymysqlreplication/binlogstream.py
```diff
@@ -134,7 +135,8 @@ def __init__(self, connection_settings, server_id, ctl_connection_settings=None,
                  freeze_schema=False, skip_to_timestamp=None,
                  report_slave=None, slave_uuid=None,
                  pymysql_wrapper=None,
-                 fail_on_table_metadata_unavailable=False):
+                 fail_on_table_metadata_unavailable=False,
+                 representation='HUMAN_READABLE'):
```
noplay: I think here I will directly put as parameter the representation class to allow anyone to implement another without touching the lib code.

noplay:
Hi,

Thanks a lot it's very interesting. What do you think about having a json method on an event to get it as a JSON? So you can use print(event.dump()) and print(event.json()) at the same time.

I think a lot of user could take advantage of your json, for example for dumping to Kafka.

pymysqlreplication/binlogstream.py
charltonaustin: Yeah that makes sense. Do you want me to make the change? I don't mind, either way.

charltonaustin:
That's a great idea. One that I wouldn't mind doing. The reason I didn't go down that path was I wanted to be able to change the configuration depending on what I was going to be doing with the data. So for instance I can pass as a parameter the expected output to a python script and then configure it that way. I don't know how other people are using this though. If you think that's better both ways make sense to me.

PR #191, opened 2017-01-28
Implement Heartbeat

baloo:
I ended up debugging a slow startup with pymysqlreplication,
the stream "stucked" for a few seconds before seeing first events coming in.
For that purpose I ended up implementing HeartbeatLogEvent and
documenting observed mysql behavior.

I also implemented the heartbeat feature, which is required for the mysqld
to filter events in gtid mode.

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -14,7 +14,7 @@ from .event import (
     QueryEvent, RotateEvent, FormatDescriptionEvent,
     XidEvent, GtidEvent, StopEvent,
     BeginLoadQueryEvent, ExecuteLoadQueryEvent,
-    NotImplementedEvent)
+    HeartbeatLogEvent, NotImplementedEvent)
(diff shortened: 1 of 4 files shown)


noplay:
Nice

baloo:
@noplay I'm concerned about the new events in the stream. This may break the client apps. Maybe I should remove it from the default set.

noplay:
In which scenario do you think it can break?

Le mar. 31 janv. 2017 à 22:55, Arthur Gautier <notifications@github.com> a
écrit :

baloo:
New unexpected event in the stream. That's the only one I see.

noplay:
If user don't want to receive unexpected events they should use the
only_events array. Perhaps we need to make that more clear.

Le mer. 1 févr. 2017 à 23:21, Arthur Gautier <notifications@github.com> a
écrit :

baloo:
Agreed

PR #197, opened 2017-03-27
Clear table_map if RotateEvent has timestamp of 0

troyjfarrell:
MySQL generates two RotateEvents when flushing binary logs.  The first
is added to the end of the old binary log.  This RotateEvent has a valid
timestamp.  The second is added to the beginning of the new binary log.
This RotateEvent has a timestamp of 0.

Under some circumstances, e.g., the MySQL server crashes, the first
RotateEvent will not be written to the binary log.  (See the [MySQL
Internals Manual 20.9](https://dev.mysql.com/doc/internals/en/event-data-for-specific-event-types.html).)

Before this change, when instantiating BinLogStreamReader with a
positive skip_to_timestamp argument, the RotateEvents with a timestamp
of 0 were ignored.  This could cause silent corruption of the table_map.

This change ensures that RotateEvents always reset the table_map,
log_pos and log_file attributes.

### DIFF
--- a/pymysqlreplication/tests/binlogfilereader.py
+++ b/pymysqlreplication/tests/binlogfilereader.py
@@ -1,0 +1,150 @@
+'''Read binlog files'''
+import struct
+
+from pymysql.util import byte2int
+from pymysqlreplication import constants
+from pymysqlreplication.event import FormatDescriptionEvent
+from pymysqlreplication.event import QueryEvent
+from pymysqlreplication.event import RotateEvent
+from pymysqlreplication.event import XidEvent
+from pymysqlreplication.row_event import TableMapEvent
+from pymysqlreplication.row_event import WriteRowsEvent
+
+class SimpleBinLogFileReader(object):
+    '''Read binlog files'''
+
+    _expected_magic = b'\xfebin'
+
+    def __init__(self, file_path, only_events=None):
+        self._current_event = None
+        self._file = None
+        self._file_path = file_path
+        self._only_events = only_events
+        self._pos = None
+
+    def fetchone(self):
+        '''Fetch one record from the binlog file'''
+        if self._pos is None or self._pos < 4:
+            self._read_magic()
+        while True:
+            event = self._read_event()
+            self._current_event = event
+            if event is None:
+                return None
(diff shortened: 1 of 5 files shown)


pymysqlreplication/tests/binlogfilereader.py
```diff
@@ -0,0 +1,150 @@
...
+            constants.XID_EVENT: XidEvent,
+            constants.TABLE_MAP_EVENT: TableMapEvent,
+            constants.WRITE_ROWS_EVENT_V2: WriteRowsEvent,
+        }.get(event.event_type)
+        return event_type in self._only_events
+
+    def _open_file(self):
+        '''Open the file at ``self._file_path``'''
+        if self._file is None:
+            self._file = open(self._file_path, 'rb+')
```
troyjfarrell: This line opens the MySQL binary log for writing.  If the user running the test does not have the necessary permissions, this will fail.

pymysqlreplication/tests/binlogfilereader.py
```diff
@@ -0,0 +1,150 @@
...
+from pymysql.util import byte2int
+from pymysqlreplication import constants
+from pymysqlreplication.event import FormatDescriptionEvent
+from pymysqlreplication.event import QueryEvent
+from pymysqlreplication.event import RotateEvent
+from pymysqlreplication.event import XidEvent
+from pymysqlreplication.row_event import TableMapEvent
+from pymysqlreplication.row_event import WriteRowsEvent
+
+class SimpleBinLogFileReader(object):
```
baloo: <3
/me sends hugs to @troyjfarrell 

That is so nice you implemented this.

pymysqlreplication/tests/binlogfilereader.py
baloo: Maybe just add a comment in `docs/development.rst`, like: 
> Run tests with:
> ```
>  py.test -k "not test_no_trailing_rotate_event"
> ```

baloo:
Failed tests looks unrelated to changes.

baloo:
@troyjfarrell may I ask you to rebase on current master (which fixed tests)?

Issue #173, opened 2016-09-22
ignored-schemas


mpieters:
So far python-mysql-replication is working fantastic, however;
a option like "ignored_schemas" whould be a welcome addition

noplay:
It's probably not big to contribute you can take example on the only_schemas  implementation. We will be happy to merge it.

baloo:
Fixed by #201

PR #201, opened 2017-04-01
Add support for ignored_tables and ignored_schema

abrarsheikh:
inspired from #173

### DIFF
--- a/pymysqlreplication/event.py
+++ b/pymysqlreplication/event.py
@@ -7,19 +7,13 @@ from pymysql.util import byte2int, int2byte
 
 
 class BinLogEvent(object):
-    def __init__(
-        self,
-        from_packet,
-        event_size,
-        table_map,
-        ctl_connection,
-        only_tables=None,
-        ignored_tables=None,
-        only_schemas=None,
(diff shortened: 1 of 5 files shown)


pymysqlreplication/event.py
```diff
@@ -7,11 +7,19 @@
...
+        event_size,
+        table_map,
+        ctl_connection,
+        only_tables=None,
+        ignored_tables=None,
+        only_schemas=None,
+        ignored_schemas=None,
+        freeze_schema=False,
+        fail_on_table_metadata_unavailable=False
+    ):
```
baloo: I'm sorry to be picky but iI kind of dislike this style, could you please avoid making the diff too large and stick to the old style (which looks pep8 compliant besides the space around the operators)?  Same comment for `binlogstream.py` and `packet.py`.

abrarsheikh:
@baloo thanks for the quick code review. Could you also bump up the package version :)

baloo:
Cc @noplay

Sadly I don't have access to the pypi myself

noplay:
I'm on it

PR #203, opened 2017-04-07
Gtid & Set functions

baloo:
Refactor closer to the spec (no conditional offset, a simpler closed-open
interval)

Implement set functions:

set - gtid, set + gtid, gtid in set, ..


noplay:
Nice :)

PR #213, opened 2017-06-12
Add implementation of read_int32

oxmane:
This method is missing, but it is being called here: https://github.com/noplay/python-mysql-replication/blob/master/pymysqlreplication/packet.py#L358

This causes the replication to break when there is a 32-bit number in a JSON column.


baloo:
Thanks for fixing my code:)

oxmane:
No problem.
Thanks @baloo !

Issue #225, opened 2017-08-29
datetime '0000-00-00' return string '0000-00-00'


seanlook:
'0000-00-00 00:00:00' is a invalid datetime but it can be inserted in `datetime` column type. If I have `f_time datetime NOT NULL` defined, the parsed value in row event is set to `None`: `row_event.py: __read_datetime2()`

Then I use all column values to insert/update, that `None` value would lead to no data matched, or insert a null value to `NOT NULL` columns(error occurs).

We all know datetime '0000-00-00' is nasty but when it already exists, we shall keep the data consistent. It's nice if a parameter like `date_tostr`  is given to `BinlogStreamReader` to decide the behavior.

I've implemented this, I will make a pull request later.

PR #228, opened 2017-09-01
fix date '0000-00-00 00:00:00.000000' parse

seanlook:
1. when there's two datetime(6) columns in table, and the first is change to invalid date like '00-00-00 00:00:00.000000'，the second parse fail
I think it's because that the first microsecond part is skipped when it returns `None`
2. Sometimes I want to convert datetime '0000-00-00 00:00:00' to exactly the string one rather than None, to keep consistent. add `date_tostr` argument to BinLogStreamReader to decide the behavior.
please refer issus #225

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -136,7 +136,8 @@ class BinLogStreamReader(object):
                  report_slave=None, slave_uuid=None,
                  pymysql_wrapper=None,
                  fail_on_table_metadata_unavailable=False,
-                 slave_heartbeat=None):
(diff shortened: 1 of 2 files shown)


dongwook-chan:
@the4thdoctor 
Hi again! I'm terribly sorry that I missed the issue you posted! Do you still think you need this feature? How would you like control over zero datetime?
1. converter feature as pymysql provides it
2. optional parameter to BinlogStreamReader

the4thdoctor:
Hi, I think an optional parameter to BinlogStreamReader would be great. It gives me the possibility of making the feature an opt in.
Ta!

PR #230, opened 2017-09-12
Feature/add general encoding

drorbar:


baloo:
While I acknowledge this is a problem the library has. I'm not sure relying on charset detection is a good way to handle this problem. I would prefer an implementation where charset choice is given to user. Should he want to rely on charset detection, fine, but as a library maintainer I'm not sure this is our responsibility.

Let me know what you think.

jasonzzz:
If column is digested like `sha1().digest()`, `chardet.detect(query)['encoding']` will be empty(utf-8 default). It will still raise UnicodeDecodeError on L191.

PR #235, opened 2017-11-09
Datetime3 read fails entire packet read

idop-alooma:
**only row_event.py changes are relevant**

**bug discription:** in some cases _fetch_one_row will result in some parsing error because of bad data packet read.
the miss read originated while reading datetime(3) rows. in cases where there was an error reading the datetime(3) from the packet the read function will return None and will skip reading the millisecond bytes, this will result in in an indentation of the packet reading, which will result error reading the rest of the packet.

**solution:** now the millisecond bytes will be trimmed in case of an error reading datetime(3) rows.


baloo:
Could you please include only the last commit in your pull request?

idop-alooma:
sorry but i'm having a lot of trouble doing it since i cant open a brunch
on your git

PR #236, opened 2017-11-16
Datetime3 read fails entire packet read

idop-alooma:
**bug description:** in some cases _fetch_one_row will result in some parsing error because of bad data packet read.
the miss read originated while reading datetime(3) rows. in cases where there was an error reading the datetime(3) from the packet the read function will return None and will skip reading the millisecond bytes, this will result in in an indentation of the packet reading, which will result error reading the rest of the packet.

**solution:** now the millisecond bytes will be trimmed in case of an error reading datetime(3) rows.

### DIFF
--- a/pymysqlreplication/row_event.py
+++ b/pymysqlreplication/row_event.py
@@ -342,6 +342,7 @@ class RowsEvent(BinLogEvent):
                 minute=self.__read_binary_slice(data, 28, 6, 40),
                 second=self.__read_binary_slice(data, 34, 6, 40))
         except ValueError:
+            self.__read_fsp(column)
             return None
         return self.__add_fsp_to_time(t, column)
 


yuvalbar:
@baloo Hey, anything missing to get this merged?

baloo:
Sorry, missed that notification

daiguadaidai:
datetime parse bug, millisecond accuracy problems。

I resolve it by edit `row_event.py`, add following code

```
212             if column.fsp == 1: microsecond *= 100000
213             elif column.fsp == 2: microsecond *= 10000
214             elif column.fsp == 3: microsecond *= 1000
215             elif column.fsp == 4: microsecond *= 100
216             elif column.fsp == 5: microsecond *= 10
```

finnal code:
```
201     def __read_fsp(self, column):       
202         read = 0           
203         if column.fsp == 1 or column.fsp == 2:
204             read = 1       
205         elif column.fsp == 3 or column.fsp == 4: 
206             read = 2       
207         elif column.fsp == 5 or column.fsp == 6: 
208             read = 3       
209         if read > 0:       
210             microsecond = self.packet.read_int_be_by_size(read)
211   
212             if column.fsp == 1: microsecond *= 100000
213             elif column.fsp == 2: microsecond *= 10000
214             elif column.fsp == 3: microsecond *= 1000
215             elif column.fsp == 4: microsecond *= 100
216             elif column.fsp == 5: microsecond *= 10
217 
218             if column.fsp % 2:
219                 return int(microsecond / 10)
220             else:
221                 return microsecond      
222 
223         return 0
```

baloo:
open a mr, comment in the issue you already opened, but don't generate noise.

Issue #185, opened 2016-12-22
Graceful exit when connect to mysql whose bin-log is not enable


cutewalker:
When connect to a mysql whose bin-log is not enable,  an error raise without any useful hint:

```
$ python dump_events.py
Traceback (most recent call last):
  File "dump_events.py", line 33, in <module>
    main()
  File "dump_events.py", line 26, in main
    for binlogevent in stream:
  File "<the-python-site-packages>/pymysqlreplication/binlogstream.py", line 371, in fetchone
    self.__connect_to_stream()
  File "<the-python-site-packages>/pymysqlreplication/binlogstream.py", line 279, in __connect_to_stream
    self.log_file, self.log_pos = cur.fetchone()[:2]
TypeError: 'NoneType' object is not subscriptable
```

From the source,  we know this is because ``show master status`` return ``Empty set``.

So, a graceful exit message or a custom exception here would be much nice.

baloo:
pull requests welcome :)

PR #245, opened 2018-01-11
Raise an explanatory exception when binary logging is not enabled. (Issue 185)

joeportela:
I was able to test by disabling binary logging on my local MySQL instance and running tests.  Certainly open to other suggestions as well. #185


noplay:
Thanks a lot

Le ven. 12 janv. 2018 à 01:12, Arthur Gautier <notifications@github.com> a
écrit :

PR #246, opened 2018-01-30
Add variable length JSON string support

mp-e:
The current implementation for reading a JSON string is incorrect and will fail to read strings longer than 127 characters. The test that I added in this branch confirms this. This correct implementation will handle longer strings according to the implementation in mysql-server and the included test passes with it.

### DIFF
--- a/pymysqlreplication/packet.py
+++ b/pymysqlreplication/packet.py
@@ -252,6 +252,22 @@ class BinLogPacketWrapper(object):
         length = self.read_uint_by_size(size)
         return self.read(length)
 
+    def read_variable_length_string(self):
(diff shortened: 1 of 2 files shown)


abrarsheikh:
looks good to me

baloo:
the code looks good, but I'd like to read find the mysql doc or code before merging this.

baloo:
sql/json_binary.cc: append_variable_length

abrarsheikh:
@baloo i would like to request a patch version bump on this package so that we can start using this in our application.

baloo:
@abrarsheikh Release 0.16 is building on travis.

baloo:
Sorry, go strait to 0.17 ^^

PR #250, opened 2018-02-08
Fix read date and datetime in case received 0000-00-00 date

almogshu:


baloo:
wouldn't a `date=Null` be a better option here?
In its current state, the code will parse `0001-01-01` the same was as `0000-00-00`.

PR #261, opened 2018-04-15
Add a stop method to stop the main loop

Photonios:
Without this, it's impossible to stop the main loop from another thread and the process is stuck forever.


baloo:
"main loop" would be in the `__iter__` if you use `BinLogStreamReader` as an iterator (github does not show comment in context if you do not change this line)

Issue #260, opened 2018-04-10
Possibly incomplete support of CRC32 checksums 


yuvalbar:
This is relating to #4 :

We have an AWS MySQL Aurora server we are reading from using this great tool. The server has CRC32 checksum turned on, and most events are read properly. However, once in a while, the Rotate event at the end of a logfile does not contain the 4 checksum bytes at the end. This causes the library to improperly read the RotateEvent, ignoring the last 4 bytes of the next binlog name:

This packet, for instance: 
``` Python
>>> x = m.fetchone()
>>> x.packet.packet._data
'\x00\x95\xda\xc7Z\x04\x917_\x0b5\x00\x00\x00\x0bd\xb3\x03\x00\x00\x04\x00\x00\x00\x00\x00\x00\x00mysql-bin-changelog.000049'
>>> x.next_binlog
u'mysql-bin-changelog.00'
```
As you can see, the packet has no checksum - it ends with the bin log file name. 
I went ahead and checked the underlying connection and the next recv returns the next packet (meaning the packet itself seems to have been read properly - it just doesn't contain a checksum.

It looks like `mysqlbinlog` reads this packet properly, which lead me to believe that this is either a bug in Aurora that causes _some_ packets to be sent w/o a checksum and the implementation of the MySQL protocol doesn't fail on it, or that this is in fact a legitimate packet structure.

The problem root cause in pymysqlreplication is here (packet.py):
``` Python
        # MySQL 5.6 and more if binlog-checksum = CRC32
        if use_checksum:
            event_size_without_header = self.event_size - 23
        else:
            event_size_without_header = self.event_size - 19
```

The code assumes that the event size should be decreased by 4 when checksum is enabled, but in this case, it shouldn't be. I haven't had time to figure out the best solution yet, so mainly raising awareness and igniting a discussion.

PR #263, opened 2018-05-09
Full handling of bin log checksums

sagiyosefia:
Following the error (#260 ), we looked into checksum handling in some other binlog reading libraries. Eventually we found this [commit](https://github.com/shyiko/mysql-binlog-connector-java/commit/1f4fe856b4a0274a8946734f130ef146462fb493) in shyiko’s mysql-binlog-connector-java repository, which seems to handle this issue.

The high-level solution here is disabling the checksum when we create a session with the db, so that all rotate events are sent from the server without a checksum. That way we can always read the next binlog file from server-auto-generated rotate events.

To handle the rotate events in the files themselves (which could have been written with a checksum even if checksum is now disabled on the server), we deduce the checksum status for each file based on the FormatDescriptionEvent in the beginning of the file instead of relying on the current server config.

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -274,10 +274,8 @@ class BinLogStreamReader(object):
         # we support it
         if self.__use_checksum:
             cur = self._stream_connection.cursor()
-            # sagi's fix
-            # cur.execute("set @master_binlog_checksum= @@global.binlog_checksum")            
             # if checksum is enabled on server - 
-            # turn it off at current session for getting rotate events always 
+            # turn it off at the current session for getting rotate events always
             # without checksum
             cur.execute("set @master_binlog_checksum= 'NONE'")
             cur.close()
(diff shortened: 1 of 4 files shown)


baloo:
I'd like to understand why mysqlbinlog doesn't complain about the missing checksums before merging this.

baloo:
It sound to me like it is related to fake rotate event which does not have checksums (because generated by master, not stored on binlog files). The correct way to handle those is to store the current log file name, and compare the log_file name in the rotateevent. If it is the same, then it's a fake rotate event and there is no checksums.

I would prefer to have an implementation similar to that instead of disabling checksums altogether.

sagiyosefia:
Hi @baloo , it looks like we have a misunderstanding. What this PR does is not turn off the checksum. It asks the server to send all auto-generated rotate events without a checksum so that we can always read them, and then takes the checksum status of each file from the format description event at the beginning of that file (which is what mysqlbinlog probably does as well). 

This is necessary because when a server checksum config is changed, previously-written rotate events in files (as well as any other event in that file) will still have/not have checksums according to the checksum configuration that was on the server when they were written. At the moment the library assumes that this config was identical to whatever is configured on the server now, which is not true.

PR #267, opened 2018-05-14
Update enum columns to support correct mysql enum indexing

idop-alooma:
When an enum value is stored in mysql it represent the index of the value (for type=enum('a','b','c') 1 represent 'a' ext... ) also, 0 represents the empty string. (when a wrong value is inserted to an enum column it will be saved as an empty string)

When the package got an enum type it changed the indexes so ('a', 'b', 'c') will be mapped to (0, 1, 2), and it transformed the values from the bin-log accordingly. So value 1 (which is 'a' in this exemp) will be changed to 0 ect..
If a value 0 was received it will be transformed to -1 which in python will be 'c'.

To solve this i added the empty string by default to an enum column metadata and removed the value transformation so that (0, 1, 2, 3) will be mapped to -> ('', 'a', 'b', 'c')

### DIFF
--- a/pymysqlreplication/column.py
+++ b/pymysqlreplication/column.py
@@ -72,7 +72,7 @@ class Column(object):
     def __read_enum_metadata(self, column_schema):
         enums = column_schema["COLUMN_TYPE"]
         if self.type == FIELD_TYPE.ENUM:
-            self.enum_values = enums.replace('enum(', '')\
+            self.enum_values = [''] + enums.replace('enum(', '')\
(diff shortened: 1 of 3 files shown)


idop-alooma:
this fixes open issue: https://github.com/noplay/python-mysql-replication/issues/258

yuvalbar:
@baloo Any estimation for merging this / any open issue we should handle?

baloo:
Sorry I missed that notification. Thanks for bringing it up again.

PR #271, opened 2018-07-04
fix column order

hkwi:
From mysql docs https://dev.mysql.com/doc/refman/5.6/en/columns-table.html,

> ORDINAL_POSITION is necessary because you might want to say ORDER BY ORDINAL_POSITION.
> Unlike SHOW, SELECT does not have automatic ordering.


baloo:
Thanks!

PR #273, opened 2018-07-10
MySQL JSON field type with string should decode in python3

### LINKED ISSUE #272: MySQL JSON field type with string should decode in python3
Table:

`CREATE TABLE test (id int, value json);`

With data:

id | value 
---- | ----
1 | {"string1":"string2"}

But event's values in python3 is :

`{"id":1,"value":{b"string1":b"string2"}`

And when use `json.dumps` will raise `TypeError: keys must be a string`

howmp:
Fix #272 

Table:

`CREATE TABLE test (id int, value json);`

With data:

id | value 
---- | ----
1 | {"string1":"string2"}

But event's values in python3 is :

`{"id":1,"value":{b"string1":b"string2"}`

And when use `json.dumps` will raise `TypeError: keys must be a string`

According to rfc7159 : [https://tools.ietf.org/html/rfc7159#section-8.1](https://tools.ietf.org/html/rfc7159#section-8.1)

> JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32.  The default
   encoding is UTF-8

So should  decode as `utf-8`

### DIFF
--- a/pymysqlreplication/packet.py
+++ b/pymysqlreplication/packet.py
@@ -266,7 +266,7 @@ class BinLogPacketWrapper(object):
             byte = byte2int(self.read(1))
             length = length | ((byte & 0x7f) << bits_read)
             bits_read = bits_read + 7
-        return self.read(length).decode()
+        return self.read(length).decode('utf-8')
 
     def read_int24(self):
         a, b, c = struct.unpack("BBB", self.read(3))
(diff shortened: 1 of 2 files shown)


howmp:
Please restart build, seems network error.

[https://travis-ci.org/noplay/python-mysql-replication/jobs/402200041](https://travis-ci.org/noplay/python-mysql-replication/jobs/402200041)

@baloo

baloo:
I'm not the owner of the project so can't restart directly. have to open/close :/ sorry

howmp:
Failed when install `cryptography` in python 3.3,But I  didn't change CI

rudaporto:
It would be good to remove support for Python 3.3 since nobody should be using it anymore (I guess).

Makes sense?

howmp:
> It would be good to remove support for Python 3.3 since nobody should be using it anymore (I guess).
> 
> Makes sense?

Yeah. I agree

Issue #288, opened 2019-04-07
Insert NULL in a boolean column returns no rows


mickey:
Hello,
Here's a test case that fails:

```python
def test_tiny_maps_to_none_2(self):
    create_query = "CREATE TABLE test (test BOOLEAN)"
    insert_query = "INSERT INTO test VALUES(NULL)"
    event = self.create_and_insert_value(create_query, insert_query)
    self.assertEqual(event.rows[0]["values"]["test"], None)
```

In this case, `event.rows` returns `[]` while `event.columns[0].name` successfully returns `test`.
Of course if I have more than 1 row, it works as demonstrated in your test case [here](https://github.com/noplay/python-mysql-replication/blob/master/pymysqlreplication/tests/test_data_type.py#L164).

mickey:
I dabbled quickly with the code and it seems that changing:

https://github.com/noplay/python-mysql-replication/blob/07e18b3c89f344db70cb968275ec0ef16d76d2cc/pymysqlreplication/row_event.py#L421-L422

to:

```python
while self.packet.read_bytes + 1 <= self.event_size:
    self.__rows.append(self._fetch_one_row())
```
See the `<` VS `<=`.

The tests are passing but I don't understand the code enough to do a PR before some approval (I'd love an explanation too - if it makes any sense of course 😬).

baloo:
I'd say this is an off by one error. I have no idea why though :x.

If you could throw a pr like:
```
 while self.packet.read_bytes < self.event_size:
    self.__rows.append(self._fetch_one_row())
```

Just removing the +1 and with your test case, that would be appreciated :)

PR #289, opened 2019-04-08
Insert NULL in a boolean column returns no rows #288

mickey:
This fixes https://github.com/noplay/python-mysql-replication/issues/288


baloo:
@mickey Thanks! Any chance you could rebase on master? I just fixed the tests, broken for an unrelated reason...

mickey:
@baloo ✅
Thanks for fixing the specs!

Issue #298, opened 2019-08-21
Json Result length not requested length


1084540622:
**When parsing MySQL 8.0 binlog using python-mysql-replication. A Value of Json field is too long.  Appearance of Result length not requested length phenomenon**

## Environment

- Python Version:  Python 3.5
- MySQL Version:  MySQL 8.0
- requirements.txt: python-mysql-replication 0.19

```txt
Traceback (most recent call last):
  File "c:\Users\zhongbiao\.vscode\extensions\ms-python.python-2019.8.30787\pythonFiles\ptvsd_launcher.py", line 43, in <module>
    main(ptvsdArgs)
  File "c:\Users\zhongbiao\.vscode\extensions\ms-python.python-2019.8.30787\pythonFiles\lib\python\ptvsd\__main__.py", line 432, in main
    run()
  File "c:\Users\zhongbiao\.vscode\extensions\ms-python.python-2019.8.30787\pythonFiles\lib\python\ptvsd\__main__.py", line 316, in run_file
    runpy.run_path(target, run_name='__main__')
  File "C:\Users\zhongbiao\AppData\Local\Programs\Python\Python35\lib\runpy.py", line 263, in run_path
    pkg_name=pkg_name, script_name=fname)
  File "C:\Users\zhongbiao\AppData\Local\Programs\Python\Python35\lib\runpy.py", line 96, in _run_module_code
    mod_name, mod_spec, pkg_name, script_name)
  File "C:\Users\zhongbiao\AppData\Local\Programs\Python\Python35\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "c:\Users\zhongbiao\Desktop\mbinlog.py", line 45, in <module>
    main()
  File "c:\Users\zhongbiao\Desktop\mbinlog.py", line 22, in main
    for row in binlogevent.rows:
  File "C:\Users\zhongbiao\AppData\Local\Programs\Python\Python35\lib\site-packages\pymysqlreplication\row_event.py", line 427, in rows
    self._fetch_rows()
  File "C:\Users\zhongbiao\AppData\Local\Programs\Python\Python35\lib\site-packages\pymysqlreplication\row_event.py", line 422, in _fetch_rows
    self.__rows.append(self._fetch_one_row())
  File "C:\Users\zhongbiao\AppData\Local\Programs\Python\Python35\lib\site-packages\pymysqlreplication\row_event.py", line 511, in _fetch_one_row
[...]

Issue #226, opened 2017-08-30
Result length not requested length


nghiaminhle:
Dear all,

Today, i've got a exception:
Result length not requested length:
Expected=8643.  Actual=6.  Position: 595.  Data Length: 601

> The exception tracing: 
> Result length not requested length:
> Expected=8643.  Actual=6.  Position: 595.  Data Length: 601
> Traceback (most recent call last):
>  
>    for row in binlogevent.rows:
>   File "/python3.6/site-packages/pymysqlreplication/row_event.py", line 428, in rows
>     self._fetch_rows()
>   File "/python3.6/site-packages/pymysqlreplication/row_event.py", line 423, in _fetch_rows
>     self.__rows.append(self._fetch_one_row())
>   File "/python3.6/site-packages/pymysqlreplication/row_event.py", line 512, in _fetch_one_row
>     row["before_values"] = self._read_column_data(self.columns_present_bitmap)
>   File "/python3.6/site-packages/pymysqlreplication/row_event.py", line 130, in _read_column_data
>     values[name] = self.__read_string(2, column)
>   File "/python3.6/site-packages/pymysqlreplication/row_event.py", line 218, in __read_string
>     string = self.packet.read_length_coded_pascal_string(size)
>   File "/python3.6/site-packages/pymysqlreplication/packet.py", line 253, in read_length_coded_pascal_string
>     return self.read(length)
>   File "/python3.6/site-packages/pymysqlreplication/packet.py", line 153, in read
>     return self.packet.read(size)
>   File "/python3.6/site-packages/pymysql/connections.py", line 250, in read
>     raise AssertionError(error)
> AssertionError: Result length not requested length:
> Expected=8643.  Actual=6.  Position: 595.  Data Length: 601

When i remove the update rows event, the problem is gone away. 
My configuration:
```
stream = BinLogStreamReader(
                connection_settings=mysql_settings,
                server_id=121
                ,only_events=[DeleteRowsEvent, WriteRowsEvent, UpdateRowsEvent]
                ,freeze_schema = False
                ,log_file='bin.002877'
                ,log_pos=753583192
                ,resume_stream = True
                ,auto_position = False
            )
```
I don't know the reason. Please help to figure out the solution for this solution

baloo:
can you share the pymysql verion and the version of pymysqlreplication?

tferreira:
I just got the same issue with WriteRowsEvent.
```
mysql-replication==0.15
PyMySQL==0.8.0
```

[...]

PR #307, opened 2019-12-13
Fix handling of JSON literal values #226 #298

juravlikk:
In case of large JSON documents, literal values should be retrieved with
`read_uint32` instead of `read_uint32` to avoid further AssertionError:

```
To compensate for the extra space needed by the redundant length
information, we will make the format allow offset and length fields to
come in two different variants: 2 bytes for documents smaller than
64KB, and 4 bytes to support larger documents.
```
https://dev.mysql.com/worklog/task/?id=8132


baloo:
Thanks!

PR #308, opened 2019-12-15
add .NET version

Oceania2018:


baloo:
There is a ton of noise in the diff with added whitespaces, I'd be happy to merge, but would you mind fixing those before?

Thank you for your contribution

Oceania2018:
Removed.

Issue #315, opened 2020-04-22
Set columns table


rspadim:
Could be possible instead of reading the column information from information schema

Read it from other table (1) or maybe from other function (2)?

Passing the function callback as parameter or column table name as string parameter?

Thanks!

For (1): https://github.com/noplay/python-mysql-replication/blob/511b42c8ac2c1682a6e2fd4d6691658245b57987/pymysqlreplication/binlogstream.py#L555

For (2): https://github.com/noplay/python-mysql-replication/blob/511b42c8ac2c1682a6e2fd4d6691658245b57987/pymysqlreplication/binlogstream.py#L543

baloo:
https://github.com/noplay/python-mysql-replication/issues/304#issuecomment-536293698

PR #316, opened 2020-04-22
Implement #315

rspadim:


baloo:
I would prefer that to be implemented like: https://github.com/noplay/python-mysql-replication/issues/304#issuecomment-536293698

PR #317, opened 2020-04-22
Allow queries to be executed at connection

rspadim:
it's usefull to set connection collate , SET and others variables


baloo:
I might be missing something here, but what is the point of setting collate on a connection that is only going to be streamed the content of the binlog?

baloo:
also, I'm pretty sure mysql server is never going to read your commands, you write them after the BINLOG_DUMP commands which have no exit conditions and are only terminated by disconnecting the reader.

PR #326, opened 2020-07-21
add PyMySQL 0.10.0 support

hkwi:
fixes includes

- pymysql version switch
- pymysql charset name conversion #325

PR #328, opened 2020-08-12
Support PyMysql with a version greater than 0.9.3

garenchan:
1. Since version 0.10.0, PyMysql removed `charset_to_encoding` method.
2. It is problematic to compare versions directly through string comparison.
> e.g. str('0.10.0') is less than str('0.6.0'), but version('0.10.0') is greater than version('0.6.0').

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -2,6 +2,7 @@
 
 import pymysql
 import struct
+from distutils.version import LooseVersion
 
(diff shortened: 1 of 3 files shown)


mattbornski:
@baloo this or #326 seem like important candidates for merge, can you review?

srikiraju:
@baloo Could we get a release tag with this fix in?

noplay:
I can push a release before the end of the week end

the4thdoctor:
Cool, I'll wait for the new release before doing a new release for pg_chameleon. 
Ta!

noplay:
Release 0.22 is out !

PR #333, opened 2020-11-06
Add logic to handle inlined ints in large json documents

mascah:
This PR fixes a bug when handling inlined ints in "large" JSON documents. When attempting to read these values, the incorrect length was used causing some cascading failures downstream trying to parse the bin logs.

Reading the correct lengths for these data types within large JSON documents resolved the issue. 

This is related to https://github.com/noplay/python-mysql-replication/pull/307

### DIFF
--- a/pymysqlreplication/packet.py
+++ b/pymysqlreplication/packet.py
@@ -394,13 +394,13 @@ class BinLogPacketWrapper(object):
             elif value == JSONB_LITERAL_FALSE:
                 return False
         elif t == JSONB_TYPE_INT16:
-            return self.read_int16()
+            return self.read_int32() if large else self.read_int16()
         elif t == JSONB_TYPE_UINT16:
-            return self.read_uint16()
+            return self.read_uint32() if large else self.read_uint16()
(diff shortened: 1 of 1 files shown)


noplay:
Thanks a lot 

Do you think it's possible to add test ? Otherwise we can merge like this

mascah:
> 
> 
> Thanks a lot
> 
> Do you think it's possible to add test ? Otherwise we can merge like this

I'd love to add some tests as well, but I ran into some snags when trying to get them to run locally. I'll take another stab at it over the next couple days.

mascah:
@noplay Any chance we could merge this in as is? I unfortunately haven't had the time to get tests written for this but I'm aware of some downstream dependencies (specifically https://github.com/singer-io/tap-mysql) that are failing on these >64KB JSON documents as well.

noplay:
What do you think @baloo ?

baloo:
I haven’t triggered the issue myself and haven't been able to experiment, but from what I read, it makes sense.

noplay:
Sorry for t he delay but new release will arrive today

PR #351, opened 2021-07-14
Support 'ZEROFILL' attribute for numeric datatypes

dongwook-chan:
I need to have the same column value on Python as it was selected on MySQL.
However the existing code ignores 'ZEROFILL' attribute.
This PR adds support for 'ZEROFILL' attribute.

1. Search for 'ZEROFILL' in column type
2. Set boolean variable zerofill for column objects
3. Format column values for columns with 'ZEROFILL' attribute

+ Added and passed data type test


noplay:
Thanks a lot I'm going to merge and release

PR #357, opened 2021-09-14
Add support for specifying an end log_pos

paulvic:
I have a use case where we want to read binary logs from a specific start position to a specific end position. 

I was originally planning to just filter the events returned by the BinLogStreamReader but it just sits waiting in `self._stream_connection.read_packet()` and never returns the events (the DB has very sporadic low traffic).

I was wondering if there would be an appetite to add support for an `end_log_pos` parameter? I've tested this out locally and it solves my problem by exiting the loop when we've moved past the end_pos.

There may be a better way to achieve my goal here, so all feedback is appreciated. Thanks!

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -130,7 +130,8 @@ class BinLogStreamReader(object):
     def __init__(self, connection_settings, server_id,
                  ctl_connection_settings=None, resume_stream=False,
                  blocking=False, only_events=None, log_file=None,
-                 log_pos=None, filter_non_implemented_events=True,
+                 log_pos=None, end_log_pos=None,
+                 filter_non_implemented_events=True,
                  ignored_events=None, auto_position=None,
                  only_tables=None, ignored_tables=None,
                  only_schemas=None, ignored_schemas=None,
@@ -152,6 +153,7 @@ class BinLogStreamReader(object):
             log_file: Set replication start log file
             log_pos: Set replication start log pos (resume_stream should be
                      true)
+            end_log_pos: Set replication end log pos
             auto_position: Use master_auto_position gtid to set position
             only_tables: An array with the tables you want to watch (only works
                          in binlog_format ROW)
@@ -205,6 +207,7 @@ class BinLogStreamReader(object):
         # Store table meta information
         self.table_map = {}
         self.log_pos = log_pos
+        self.end_log_pos = end_log_pos
         self.log_file = log_file
         self.auto_position = auto_position
         self.skip_to_timestamp = skip_to_timestamp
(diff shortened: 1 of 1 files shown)


dongwook-chan:
This is an interesting feature.
I do have some questions, though.
What behavior would you expect when all the events in the current binlog file is exhausted?
Go on to the next binlog file and begin from log_pos 0? Or just stop at it?

paulvic:
> This is an interesting feature.
> I do have some questions, though.
> What behavior would you expect when all the events in the current binlog file is exhausted?
> Go on to the next binlog file and begin from log_pos 0? Or just stop at it?

Thanks for the reply -  good questions! For my use case, I would always be passing a start and end position in the same file. If the current binlog file is exhausted, then I think perhaps it would be best to just stop there.

paulvic:
@dongwook-chan any thoughts on how to proceed? I can add an `end_log_file` parameter if you think that's a good idea.

dongwook-chan:
@paulvic 
Great idea.

Before going any further, I need to take a look at code.
I'll leave comments if anything comes across my mind.

dongwook-chan:
@paulvic 
Sorry for the delay.
It was Korean thanks giving holiday.
I've studied code for a bit.
It seems that the main issue will be when to stop the stream.
It might take a while to determine when.
I'll leave comments for any progress.

dongwook-chan:
@paulvic
Seems like you made some changes in paulvic:patch-1 for your on sake.
Some of the changes you made are not eligible to be merged in to the project's master branch.
So I don't think I can continue on with your forked repo anymore.

Plus, I think using flag variable to 'should_continue' isn't the correct solution.
I would like to know if it is tested.
Do you want me continue the work on my forked repo?

dongwook-chan:
@paulvic 
I removed `should_continue` variable and added test in my forked branch.
However, I think you should be credited for your contribution.
There seems to be a few options:

* Remove commits in `paulvic:patch-1` that are not relevant to this PR. (Or maybe move the commits to another branch of yours?)
* Grant me write permission to intercom:patch-1

I'll be waiting for your reply before making any progress with the PR.

dongwook-chan:
@paulvic 
I would like to know if you want `end_log_pos` to be inclusive.
I assumed that you wanted so and the last event to be read will be binlog event that ends at end_log_pos.

paulvic:
> @paulvic I removed `should_continue` variable and added test in my forked branch.
Hi @dongwook-chan, thanks for your help here. I added a comment to your proposed changes https://github.com/dongwook-chan/python-mysql-replication/commit/67d3a71e123563ccfbfa277e3be68955081c3b2d.

> * Remove commits in `paulvic:patch-1` that are not relevant to this PR. (Or maybe move the commits to another branch of yours?)
Which commits are you referring to? There are just 3 commits, and I believe they're all related to this PR but maybe I'm misunderstanding 😄.

> I would like to know if you want `end_log_pos` to be inclusive.
Yes 👍 , I would like the `end_log_pos` to be inclusive.

paulvic:
> Hi @paulvic I'm no longer very active but I'm really excited to see intercom is using my code for something :D

😄 @noplay We're appreciative your work here!

paulvic:
> @paulvic
> 
> I've fixed the issue. I have done some tests and it seems to work right. Would you please test the code for your specific case? I'll ask noplay, the owner to merge this PR if no issue arises.

@dongwook-chan - confirmed the fix works for me 👍

PR #360, opened 2021-09-24
Status vars

dongwook-chan:
Status variables contain useful information:  
1. List of databases affected by the query
2. Runtime value of system variables (SQL_MODE, AUTOCOMMIT, CHARSET_SERVER, etc.)
3. timezone of the master

Users of this library should benefit from new information.  
However, if anyone raises performance issues regarding the logic,  
this logic can revised to be provided selectively depending on a newly added option.   

A status variable in query events is a sequence of status KEY-VALUE pairs.
1. A KEY is parsed first in `pymysqlreplication.event.QueryEvent.__init__()`.
2. Respective VALUE is parsed in new method `pymysqlreplication.event.QueryEvent._read_status_vars_value_for_key()`.
3. This repeats until all KEY-VALUE pairs are exhausted.
Added and passed tests.

For further information please refer to;  
* In [MySQL Documentation](https://dev.mysql.com/doc/),
    * [Syntax of system variable field in query event](https://dev.mysql.com/doc/internals/en/query-event.html#q-microseconds)
* In [mysql-server Source Code](https://github.com/mysql/mysql-server), 
    * [Definition of enumeration for system variable keys](https://github.com/mysql/mysql-server/blob/beb865a960b9a8a16cf999c323e46c5b0c67f21f/libbinlogevents/include/statement_events.h#L463-L532)
    * [Semantics of system variable values](https://github.com/mysql/mysql-server/blob/beb865a960b9a8a16cf999c323e46c5b0c67f21f/libbinlogevents/include/statement_events.h#L156-L448)

### DIFF
--- a/pymysqlreplication/constants/STATUS_VAR_KEY.py
+++ b/pymysqlreplication/constants/STATUS_VAR_KEY.py
@@ -11,7 +11,7 @@
     mysql-server: https://github.com/mysql/mysql-server/blob/beb865a960b9a8a16cf999c323e46c5b0c67f21f/libbinlogevents/include/statement_events.h#L463-L532
     MySQL Documentation: https://dev.mysql.com/doc/internals/en/query-event.html
 
-    From mysql-server source code, edited buy dongwook-chan
+    Status variable key names From mysql-server source code, edited by dongwook-chan
 """
 
 # KEY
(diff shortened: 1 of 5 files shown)


noplay:
Sorry for the delay will look in details next week

dongwook-chan:
Nope, I've tested the new feature several times but not yet.
If any issue rises regarding the feature,
I'll add a parameter to BinLogStreamReader.__init__() so that users could choose whether to use this feature.

PR #363, opened 2021-12-02
improve performance and documentation, skip the first 2 events after reconnected

woods-chen:
* sleep 5s after disconnected accidently and skip the first 2
      events after reconnected
    * symplify BinLogStreamReader.fetchone
    * improve documentation of BinLogStreamReader
    * some cleanup and compacting works
    * tests/test_data_type.py: fix syntax errors
    * tests/test_basic.py: dynamic get an end position instead a fixed one

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -2,6 +2,7 @@
 
 import pymysql
 import struct
+import time
 from distutils.version import LooseVersion
 
 from pymysql.constants.COMMAND import COM_BINLOG_DUMP, COM_REGISTER_SLAVE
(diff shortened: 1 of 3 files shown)


dongwook-chan:
@woods-chen 
I love the `* symplify BinLogStreamReader.fetchone` feature.
I think it will greatly improve the performance module.

However, I'm a little concerned about `* sleep 5s after disconnected accidently and skip the first 2
  events after reconnected` feature.
The very first event the module receives upon connection is `RotateEvent`, which triggers `table_map` to be flushed:
https://github.com/noplay/python-mysql-replication/blob/7fd706d1686da90cdb3991279c764f057e24a1f6/pymysqlreplication/binlogstream.py#L507-L519
If reboot of MySQL caused disconnection, mysql-replication will move on without flushing `table_map` to be filled with table ids that is not valid any more after reboot.

Nice work with the optimization, but I think we should keep parsing the first 2 events for maintaining valid `table_map`.

dongwook-chan:
@woods-chen 
Thank you for the correction.

PR #374, opened 2022-04-16
Fix RowsEvent.__is_null usage in tests

oseemann:
The test `TestDataType::test_null_bitmask` currently fails with
```
AttributeError: type object 'RowsEvent' has no attribute '_TestDataType__is_null'`
```

The test method in the `TestDataType` class calls the `__is_null` method of the
`RowsEvent` class. Due to name mangling of "private" methods/variables in
Python [1] the calling class name is prepended to the method name. This makes
the call fail when called from a foreign class.

This commit fixes that by
a) Renaming `__is_null` to `_is_null` which is only "private" by convention and
   involves no name mangling.
b) Makes the method a staticmethod, as it's not depending on any instances or
   class attributes/ methods, and is referenced by class only (and not by
   instance) in the test.

[1] https://docs.python.org/3/tutorial/classes.html#private-variables

### DIFF
--- a/pymysqlreplication/row_event.py
+++ b/pymysqlreplication/row_event.py
@@ -85,7 +85,8 @@ class RowsEvent(BinLogEvent):
             if self._fail_on_table_metadata_unavailable:
                 raise TableMetadataUnavailableError(self.table)
 
-    def __is_null(self, null_bitmap, position):
+    @staticmethod
+    def _is_null(null_bitmap, position):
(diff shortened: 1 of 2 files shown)


oseemann:
@julien-duponchelle Can you have a quick look at this please, and also https://github.com/julien-duponchelle/python-mysql-replication/pull/375? 
I have a few more changes lined up, but I'd like most of the existing tests to pass first, before adding more.

PR #376, opened 2022-04-21
Run tests as GitHub Actions

oseemann:
To easier find out when changes break the tests.

- Use Python 3.7 for starters, because it's the oldest still-supported CPython
  interpreter version. More can be added in a subsequent step.
- Use "docker compose" to run the Percona versions from the dev readme. Different
  MySQL implementations (e.g. MariaDB, Oracle, etc.) could be added on top.
- The test suite is skipping `test_no_trailing_rotate_event` because it requires a
  different test setup, with access to the binlog files.
- The test suite also skips `test_end_log_pos` for now, because it currently
  fails and the best solution wasn't obvious to me, the original author should
  have a look.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,0 +1,33 @@
+name: PyTest
+on: push
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    timeout-minutes: 2
+
+    steps:
+      - name: Check out code
+        uses: actions/checkout@v2
+
+      - name: Setup Python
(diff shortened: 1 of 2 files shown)


julien-duponchelle:
Thanks a lot that a good improvement !

julien-duponchelle:
Do I need d to do something to enable it ?

oseemann:
No it works right away, once merged:
https://github.com/julien-duponchelle/python-mysql-replication/runs/6124868470?check_suite_focus=true

Issue #311, opened 2020-02-24
Json type 150 is not handled


hgwat:
Received this error whilst listening to binlogs in AWS RDS:

`Traceback (most recent call last):
  File "/usr/local/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/usr/local/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "./main.py", line 25, in startDBListener
    "secret": gfire
  File "/usr/src/app/database_listener.py", line 47, in listen
    for row in binlogevent.rows:
  File "/usr/local/lib/python3.7/site-packages/pymysqlreplication/row_event.py", line 427, in rows
    self._fetch_rows()
  File "/usr/local/lib/python3.7/site-packages/pymysqlreplication/row_event.py", line 422, in _fetch_rows
    self.__rows.append(self._fetch_one_row())
  File "/usr/local/lib/python3.7/site-packages/pymysqlreplication/row_event.py", line 511, in _fetch_one_row
    row["before_values"] = self._read_column_data(self.columns_present_bitmap)
  File "/usr/local/lib/python3.7/site-packages/pymysqlreplication/row_event.py", line 181, in _read_column_data
    values[name] = self.packet.read_binary_json(column.length_size)
  File "/usr/local/lib/python3.7/site-packages/pymysqlreplication/packet.py", line 352, in read_binary_json
    return self.read_binary_json_type(t, length)
  File "/usr/local/lib/python3.7/site-packages/pymysqlreplication/packet.py", line 385, in read_binary_json_type
    raise ValueError('Json type %d is not handled' % t)
ValueError: Json type 150 is not handled`

sumanau7:
@hgwat Do you have a reproducible example ?

SriSivaC:
> Received this error whilst listening to binlogs in AWS RDS:
> 
> ```
> Traceback (most recent call last):
>   File "/usr/local/lib/python3.7/threading.py", line 917, in _bootstrap_inner
>     self.run()
>   File "/usr/local/lib/python3.7/threading.py", line 865, in run
>     self._target(*self._args, **self._kwargs)
>   File "./main.py", line 25, in startDBListener
>     "secret": gfire
>   File "/usr/src/app/database_listener.py", line 47, in listen
>     for row in binlogevent.rows:
>   File "/usr/local/lib/python3.7/site-packages/pymysqlreplication/row_event.py", line 427, in rows
[...]

PR #377, opened 2022-04-24
Fix decoding large json arrays

oseemann:
Without the fix, the included test fails with

```
AssertionError: Result length not requested length:
Expected=2.  Actual=0.  Position: 500079.  Data Length: 500079
```
But I suspect (and have seen) also other errors, depending on specific contents
of the byte stream, e.g. ("Json type $random is not handled").

Int32/UInt32 values in json arrays are encoded inline in the "large" encoding
mode, but they are stored using 4 bytes, not 8.  So the `read_uint64()` calls
are wrong here, `read_uint32()` works. Also the check for `large` in that case
is redundant, as `large` is always True for inlined int32/uint32 values.

Relevant parts in mysql server code are functions `serialize_json_array`,
`should_inline_value` and `insert_offset_or_size` in `sql/json_binary.cc` [1].

I believe this fixes issue #311.

[1] https://github.com/mysql/mysql-server/blob/5.7/sql/json_binary.cc

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,5 +1,5 @@
 name: PyTest
-on: push
+on: [push, pull_request]
 
 jobs:
   test:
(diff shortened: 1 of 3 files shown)


julien-duponchelle:
Thank you so much I'm going to merge and look to release a new version

PR #378, opened 2022-04-29
Run tests with different Python versions

oseemann:
Use a matrix strategy to run the test suite with different Python interpreter
versions.


julien-duponchelle:
Perfect !

PR #387, opened 2022-11-27
fix: redis cache example

fkromer:
Fix `TypeError: BinLogStreamReader.__init__() missing 1 required positional argument: 'server_id'` in redid cache example.


julien-duponchelle:
Thanks!

PR #390, opened 2022-12-18
Unit testing for GTID/GtidSet

darnuria:
It's a bit redundant with test about GtidSet but if it makes sense I can add more.

PR #391, opened 2022-12-18
Avoid reparsing a Gtid to copy in sub/add

darnuria:
Wanted to avoid some reparsing in `sub`, `add`.


darnuria:
Reduced a lot the scope of the PR :)

darnuria:
Would be safer to land after #390, or I just missed test for add/sub.

PR #394, opened 2022-12-24
Gtid: feature gate __cmp__ since it's python2.

darnuria:
I can also remove it if python 2.7 support is no longer needed.

### DIFF
--- a/pymysqlreplication/gtid.py
+++ b/pymysqlreplication/gtid.py
@@ -281,11 +281,6 @@ class Gtid(object):
             else '%d' % x
             for x in intervals])))
 
-    def __cmp__(self, other):
-        if other.sid != self.sid:
-            return cmp(self.sid, other.sid)
-        return cmp(self.intervals, other.intervals)
-
     def __eq__(self, other):
         if other.sid != self.sid:
             return False


pymysqlreplication/gtid.py
```diff
@@ -246,6 +247,10 @@ def decode(cls, payload):
             for x in intervals])))
 
     def __cmp__(self, other):
+        """Compare two Gtid only work in python 2.7 since it rely on cmp."""
+        if version_info[0] >= 3:
+            raise NotImplemented
+
```
darnuria, line 253: Found while doing typing with mypy:

```
pymysqlreplication/gtid.py:254: error: Name "cmp" is not defined  [name-defined]
pymysqlreplication/gtid.py:255: error: Name "cmp" is not defined  [name-defined]
```

julien-duponchelle:
Oh nice!

darnuria:
@julien-duponchelle so it raise one big question: what can I do? Remove if, if python2 is dropped I think it's for the best (not an expert in python). Or I can feature gate it (may need a test to avoid footgunning) but I don't have a python2 anymore. XD

julien-duponchelle:
I'm happy to drop python 2 !

darnuria:
> I'm happy to drop python 2 !

Done, sorry for the lag busy days! ;)

julien-duponchelle:
Sorry for the late response but I merged it and I'm in the process of doing a release

PR #395, opened 2023-01-05
Add support for ignoring string decode errors

paulvic:
We discovered that some rows in our mysql db contain incorrectly encoded utf8 strings and when iterating through the binary logs, we see the following error when decoding these values:
```
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xed in position 29: invalid continuation byte
```

We'd like to be able to ignore such errors when decoding.

This change introduces a new optional parameter for the `BinLogStreamReader` called `ignore_decode_errors`. When set to True, the malformed data will be ignored.

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -140,7 +140,8 @@ class BinLogStreamReader(object):
                  pymysql_wrapper=None,
                  fail_on_table_metadata_unavailable=False,
                  slave_heartbeat=None,
-                 is_mariadb=False):
+                 is_mariadb=False,
+                 ignore_decode_errors=False):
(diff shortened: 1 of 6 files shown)


paulvic:
@julien-duponchelle please let me know what you think about this approach. Thanks 🙇

julien-duponchelle:
Sorry for the very late answer and thank you !!!

thiagoschonrock:
Hy @julien-duponchelle can you tell me if this merge has been updated on https://pypi.org/ ?

julien-duponchelle:
> Hy @julien-duponchelle can you tell me if this merge has been updated on https://pypi.org/ ?

yes

PR #409, opened 2023-07-25
Fix test code - 'isMariaDB'

cucuridas:
The result value executed through the fechone() function is passed in the form of a tuple, but the syntax to compare isMariaDB function always has an error that returns False because it checks for string type data
* before code block (pymysqlreplication/tests/base.py)
``` 
    def isMariaDB(self):
        if self.__is_mariaDB is None:
            self.__is_mariaDB = "MariaDB" in self.execute("SELECT VERSION()").fetchone()
        return self.__is_mariaDB
```
* image
image

* fixed
Change to see the first element in the tuple lookup list

### DIFF
--- a/pymysqlreplication/tests/base.py
+++ b/pymysqlreplication/tests/base.py
@@ -67,7 +67,7 @@ class PyMySQLReplicationTestCase(base):
 
     def isMariaDB(self):
         if self.__is_mariaDB is None:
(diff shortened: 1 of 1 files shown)


cucuridas:
I need to fix it, so I'll close it

cucuridas:
I changed the contents and opened it again

PR #411, opened 2023-07-28
Fix test code - 'isMariaDB'

cucuridas:
The result value executed through the fechone() function is passed in the form of a tuple, but the syntax to compare isMariaDB function always has an error that returns False because it checks for string type data
* before code block (pymysqlreplication/tests/base.py)
``` 
    def isMariaDB(self):
        if self.__is_mariaDB is None:
            self.__is_mariaDB = "MariaDB" in self.execute("SELECT VERSION()").fetchone()
        return self.__is_mariaDB
```
* image
image

* fixed
Change to see the first element in the tuple lookup list


dongwook-chan:
@cucuridas 
Holy shit...
You should probably close this PR.
You have 100 commits in total most of which are irrelevant to this.

cucuridas:
Ooops.... Sory I will change this

julien-duponchelle:
It's the wrong target branch need to be main. Sorry for that.

Le ven. 28 juil. 2023 à 16:13, cucuridas ***@***.***> a
écrit :

dongwook-chan:
@julien-duponchelle 
Thank you for reviewing this PR. @cucuridas is a good friend of mine and also one of the mentees I told you earlier. My mentees will submit quite a few PRs during the next couple of months. I'll make sure the PRs thoroughly reviewed in advance and something like this won't happen again :) Thank you again for your hard work on this repo!

PR #416, opened 2023-08-12
Feature : rows query log event

mirageoasis:
- first we added keyword arguments in setup function for more flexibility
- second we implemented yet unimplemented event RowsQueryLogEvent
- finally we wrote test code of what we implemented

image
- this documentation states that binlog format will be deprecated which means our pr will be important


mirageoasis:
fixed codes which caused conflicts!

PR #418, opened 2023-08-14
refactor : refactor loop inside _read_column_data method

mirageoasis:
* refactored structure inside loop to improve readability

* separated read values[name] and nullBitmapIndex increase because I think reading and chaning value should not be done once

image

* refactored nullBitmapIndex to null_bitmap_index to follow pep8


julien-duponchelle:
I'm sorry but the code will need to be rebased against the last version

mirageoasis:
Done with conflicts!

PR #420, opened 2023-08-14
Fix typo in XAPrepareEvent comment

heehehe:
There is a typo in the comment of XAPrepareEvent (contans → contains)


julien-duponchelle:
Thanks for the fix

PR #424, opened 2023-08-14
Fix typo in bitmap.py

why-arong:
There is a minor typo in the annotation for the BitCount function on bitmap.py.
It's trivial, but I'd appreciate it if you could consider it
tot**o**l -> total


julien-duponchelle:
Totol 😅

PR #426, opened 2023-08-14
Fix github action workflow

lre12:
## 1. Fix github action workflow
- connection check with mysql client in github actions
image

## 2. ADD docker-compose-test.yml and test.Dockerfile for local test
- Added to run test code at local


julien-duponchelle:
Thanks a lot !

PR #427, opened 2023-08-15
Fix typo in row_event.py

mjs1995:
This PR addresses a typo found in the row_event.py.

this a dirty hack -> this `is` a dirty hack


julien-duponchelle:
Thanks

Issue #430, opened 2023-08-23
Bug - Increase in Innodb_history_list_length


davinc71998:
## Problem
MySQL 8.0+: The value of Innodb_history_list_length steadily increases after running the program
The increase in Innodb_history_list_length is expected to have the following effects on MySQL
* Increased response time to user query

## Bug description
https://github.com/julien-duponchelle/python-mysql-replication/blob/62a009cd81b3a0c4928933cf4d8657376bfee5b1/pymysqlreplication/binlogstream.py#L638-L655
According to the logic written in binlogstream.py
After executing a query through the cursor, the transaction remains in place until the program exits

![스크린샷 2023-08-22 오후 11 06 03](https://github.com/julien-duponchelle/python-mysql-replication/assets/65767734/285f226d-eeba-416d-85f5-473811972f03)

the transaction is being maintained, the Innodb_history_list_length value will continue to increase when DML is performed.
![KakaoTalk_Photo_2023-08-22-23-18-38](https://github.com/julien-duponchelle/python-mysql-replication/assets/65767734/d24cb198-54e5-4298-9369-0d9e33c92aa6)

## Solution
After executing the query, ``COMMIT``  the transaction so that it can be terminated normally.

dongwook-chan:
I'm glad that this issue was created and hope this issue is immediately attended. I work at Kakao which operates the most prominent messaging service in South Korea. My team integrated the python-mysql-replication for a range of applications. However, given the substantial DML traffic we handle (we have about 48m monthly active users), we observed a surge in `Innodb_history_list_length`, leading to challenges like delayed query responses and heightened server load.

PR #431, opened 2023-08-23
fix - Bug/increase history list length

cucuridas:
### Overview
issue number #430

When using the `pymysqlreplication` package, there was a recurring issue with the `history list length` value under the following circumstances:

- MySQL version 8.x or higher.
- Continuous execution of DML (Data Manipulation Language) statements.

**This problem led to a gradual slowing down of SELECT queries.**

For detailed information, you can refer to the following reference:

https://minervadb.xyz/troubleshooting-innodb-history-length-with-hung-mysql-transaction/

As a result, I investigated the problem area and made necessary modifications. After making the changes, you were able to compare the performance before and after the fix, using monitoring tools such as Prometheus and MySQL Exporter.

### Fixed

    
- **Modified code**
    
    ```python
    def __get_table_information(self, schema, table):
            for i in range(1, 3):
                try:
                    if not self.__connected_ctl:
                        self.__connect_to_ctl()
    
                    cur = self._ctl_connection.cursor()
                    cur.execute("""
                        SELECT
                            COLUMN_NAME, COLLATION_NAME, CHARACTER_SET_NAME,
                            COLUMN_COMMENT, COLUMN_TYPE, COLUMN_KEY, ORDINAL_POSITION,
                            DATA_TYPE, CHARACTER_OCTET_LENGTH
                        FROM
                            information_schema.columns
                        WHERE
    			  table_schema = %s AND table_name = %s
                        ORDER BY ORDINAL_POSITION
                        """, (schema, table))
                    result = cur.fetchall()
                    cur.close()
    
                    return result
    								...
    								...
    ```
    

### Performance Comparison After the Code Modifications

- Before modification </br>
 image   
 
- After modification </br>    
 image

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -256,6 +256,7 @@ class BinLogStreamReader(object):
             self._ctl_connection_settings = dict(self.__connection_settings)
         self._ctl_connection_settings["db"] = "information_schema"
         self._ctl_connection_settings["cursorclass"] = DictCursor
+        self._ctl_connection_settings["autocommit"] = True
         self._ctl_connection = self.pymysql_wrapper(**self._ctl_connection_settings)
         self._ctl_connection._get_table_information = self.__get_table_information
         self.__connected_ctl = True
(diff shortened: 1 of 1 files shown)


dongwook-chan:
This PR makes it feasible for major corporations handling extensive traffic to utilize this library. It addresses potential query response lags and reduces the strain on database servers. One of the standout features of the replication protocol is its **minimal impact on database performance**, and this PR ensures the library capitalizes on that strength.

julien-duponchelle:
Great , thanks for the investigation  . Small fix but big  problem indeed

Issue #432, opened 2023-08-23
Use simplified super() calling in classes


soulee-dev:
## Issue Description

With the introduction of Python 3, the `super()` function can be called without any arguments in instance methods, automatically determining the class and self. This helps in making the code cleaner and more maintainable.

### Current Code
The current usage of `super()` in the classes is as follows:

```python
super(ClassName, self).__init__(from_packet, event_size, table_map)
```

### Proposed Change
The new and more concise usage, available from Python 3 onwards, should be applied:

```python
super().__init__(from_packet, event_size, table_map)
```

### Benefits
- Code readability: Simplifying the call to `super()` makes the code easier to read.
- Code maintainability: If the class name changes in the future, you won't have to update the call to `super()`.

### Suggested Implementation

I can make a pull request implementing these changes across the codebase, if that's acceptable.

PR #433, opened 2023-08-23
Refactor super() usage in classes

soulee-dev:
## Description

This PR addresses the use of the `super()` function in the classes. Starting with Python 3, `super()` can be called without arguments in instance methods, automatically determining the class and self. This PR applies this simplification, enhancing the code's readability and maintainability.

### Changes
- Replaced all occurrences of `super(ClassName, self).__init__` with the simplified `super().__init__` in the classes.

### Before
```python
super(ClassName, self).__init__(from_packet, event_size, table_map) ...
```

### After
```python
super().__init__(from_packet, event_size, table_map) ...
```

### Benefits
- **Readability**: The new syntax is cleaner and more in line with modern Python practices.
- **Maintainability**: If the class name changes in the future, there's no need to update the call to `super()`.

### Related Issue
Fixed: #432

### DIFF
--- a/pymysqlreplication/event.py
+++ b/pymysqlreplication/event.py
@@ -56,7 +56,7 @@ class GtidEvent(BinLogEvent):
     """GTID change in binlog event
     """
     def __init__(self, from_packet, event_size, table_map, ctl_connection, **kwargs):
-        super(GtidEvent, self).__init__(from_packet, event_size, table_map,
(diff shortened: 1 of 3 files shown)


dongwook-chan:
The modifications have greatly boosted the readability of the code. It's not just clearer but also more concise.

julien-duponchelle:
Great

Issue #445, opened 2023-08-25
increasing handler_read_rnd


ebang091:
# Problem
Detected increasing handler_read_rnd.
<img width="917" alt="스크린샷 2023-08-25 오후 4 33 04" src="https://github.com/julien-duponchelle/python-mysql-replication/assets/108053947/66459223-d150-4fd5-afa2-bc54987b941c">

# Concerns
This can lead to an increase in I/O operations, as the system needs to read through potentially large amounts of data. This can strain the disk subsystem and lead to increased latency for other operations.
Also, Sequentially reading through rows can lead to increased CPU usage, especially if the dataset is large.

# Solution
This can be done by removing 'ORDER BY' statement in table_map event.

<img width="917" alt="스크린샷 2023-08-25 오후 4 43 56" src="https://github.com/julien-duponchelle/python-mysql-replication/assets/108053947/5d29ce7b-d6ba-45ab-970d-980c81f431f5">

PR #448, opened 2023-08-26
Optimize `Handler_read_rnd` by removing ORDER BY clause

soulee-dev:
# Problem
We observed that the `Handler_read_rnd` was continually increasing whenever `__get_table_information` was called upon starting PyMyRepl. The reason was found to be the usage of the `ORDER BY` clause on a non-indexed column within the function.

# Solution
Since the table entries were empty, and `ORDER BY` was being used on a non-indexed column, this led to an increase in the `Handler_read_rnd`. As a remedy, we removed the `ORDER BY` clause from the SQL query and instead used Python's `sort()` method to sort the results. Since `ORDINAL_POSITION` is auto-incremental, this change will return the same sorted results as the original `ORDER BY` clause, without the associated performance penalty.

# Changes
Removed the `ORDER BY` clause from the SQL query within `__get_table_information`.
Implemented sorting using Python's `sort()` method based on `ORDINAL_POSITION`.

# Benefits
Reduces the overhead associated with using the `ORDER BY` clause on a non-indexed column.
Mitigates the increase in `Handler_read_rnd`, leading to improved performance.

Fixed: #445

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -651,7 +651,6 @@ class BinLogStreamReader(object):
                         information_schema.columns
                     WHERE
                         table_schema = %s AND table_name = %s
-                    ORDER BY ORDINAL_POSITION
                     """, (schema, table))
                 result = cur.fetchall()
                 cur.close()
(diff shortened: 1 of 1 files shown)


julien-duponchelle:
That's great !

PR #451, opened 2023-08-26
PyMyRepl is featured on AWS

soulee-dev:
PyMyRepl is featured on AWS
[Streaming Changes in a Database with Amazon Kinesis
](https://aws.amazon.com/blogs/database/streaming-changes-in-a-database-with-amazon-kinesis/)


soulee-dev:
Seems like commit history is broken.
I'll try to fixit.

PR #452, opened 2023-08-26
PyMyRepl is featured on AWS

soulee-dev:
https://aws.amazon.com/blogs/database/streaming-changes-in-a-database-with-amazon-kinesis/


julien-duponchelle:
Great finding !

PR #454, opened 2023-08-27
implemented previous_gtid_event

suhwan-cheon:
### Overview
- This PR implements mysql's previous_gtid_log_event [docs](https://dev.mysql.com/doc/dev/mysql-server/latest/classPrevious__gtids__log__event.html)
- It is used to record the gtid_executed in the last binary log file, for ex after flush logs, or at the starting of the binary log file

**Development Content**
- Created the 'PreviousGtidEvent' class


julien-duponchelle:
Thanks a lot

PR #458, opened 2023-08-28
Enhance Data Integrity with Binlog Event Checksum Verification 

why-arong:
## Overview
This PR add supports for verifying event checksums when the [binlog_checksum](https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_binlog_checksum) option is enabled in MySQL/MariaDB([5.7,](https://dev.mysql.com/doc/refman/5.7/en/replication-options-binary-log.html#sysvar_binlog_checksum) [8.0](https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_binlog_checksum)). It enhances data integrity and security by validating each event in the binary log.

## Changes
Key updates in this PR include:
- Introduction of the _verify_event method in the BinlogEvent class. When both the [binlog_checksum](https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_binlog_checksum) option of MySQL/MariaDB and the verify_checksum option of BinLogStreamReader are active, this function:
   - Reads data, omitting the final CRC32 4 bytes.
   - Applies the [zlib CRC32 algorithm](https://github.com/madler/zlib/blob/master/crc32.c) on the data, then contrasts the outcome with the last 4 CRC32 bytes.
   - Restores the packet's state to its original condition post-execution.

The reason this method is effective lies in the shared use of the [CRC32 algorithm by both MySQL and this method](https://dev.mysql.com/doc/dev/mysql-server/latest/namespacebinary__log.html#a408d6e36390dacc37eb41f7459720e3ca0e7fe0f8d356d43a9c59280e0900d9df).

## Test Code
To ensure the accuracy of the _verify_event method, test code has been designed to handle two main scenarios: valid events and corrupted events. The respective tests can be accessed in [this source code file](https://github.com/julien-duponchelle/python-mysql-replication/blob/a2896ac0e808c9cc00135139181a43647d4cef19/pymysqlreplication/tests/test_basic.py#L529C4-L576C1).

## Benefits

- Enhances data integrity by verifying event checksums.
- Improves system reliability and reduces the risk of data loss.

### Comment
Slave already has these features. When [slave_sql_verify_checksum](https://dev.mysql.com/doc/refman/8.0/en/replication-options-replica.html#option_mysqld_slave-sql-verify-checksum) is enabled, the replication SQL thread uses checksums read from relay logs to verify data. I believe that incorporating these functionalities into our project would be highly advantageous.


julien-duponchelle:
That's great !

PR #460, opened 2023-08-31
Developing Logging Feature for 'BinLogStreamReader' Object Creation Parameters

cucuridas:
### Overview
A functionality has been developed to log the parameters provided by the user when creating a BinLogStreamReader object.
<br>
 This logging feature is implemented to track the input parameters used to create the object, and it is designed to work within the context of a larger package that might be imported and utilized by other programs. 
Therefore, specific log configurations, formats, or handling methods have not been implemented. Instead, this feature has been designed to log messages at the INFO level for visibility.

The parameter 'enable_logging' allows you to check the settings of the BinlogStreamReader object. By default, this parameter is set to 'True'. If you want to disable logging, you can simply set its value to 'False'.

</br>
</br>

* **The log object can be instantiated and utilized as demonstrated in the `dump_events.py` file as follows**
 ```
def main():
    log = logging.getLogger()
    log.setLevel(logging.INFO)
    
    stream = BinLogStreamReader(connection_settings=MYSQL_SETTINGS,
                                server_id=3,
                                blocking=True)

    for binlogevent in stream:
        binlogevent.dump()

    stream.close()
```

* **Result**
```
INFO:root:Set parameter - connection_settings, value - {'host': '127.0.0.1', 'port': 3306, 'user': 'root', 'passwd': '', 'charset': 'utf8'}
INFO:root:blocking is True
INFO:root:allowed_events: [StopEvent, ExecuteLoadQueryEvent, TableMapEvent, FormatDescriptionEvent, BeginLoadQueryEvent, UpdateRowsEvent, QueryEvent, XAPrepareEvent, DeleteRowsEvent, WriteRowsEvent, RotateEvent, MariadbGtidEvent, HeartbeatLogEvent, GtidEvent, XidEvent]
INFO:root:allowed_events_in_packet: [StopEvent, FormatDescriptionEvent, XAPrepareEvent, RotateEvent, MariadbGtidEvent, GtidEvent, ExecuteLoadQueryEvent, TableMapEvent, BeginLoadQueryEvent, UpdateRowsEvent, QueryEvent, DeleteRowsEvent, WriteRowsEvent, HeartbeatLogEvent, XidEvent]
```

### Benefits
This functionality provides a clear visibility into "What parameters did the user provide when creating a 'BinLogStreamReader' object?" The logging messages help users understand the configuration and input parameters utilized during the object creation process.

### CODE REVIEW

File: pymysqlreplication/binlogstream.py [hunk 1]
```diff
@@ -142,7 +143,8 @@ def __init__(self, connection_settings, server_id,
                  slave_heartbeat=None,
                  is_mariadb=False,
                  annotate_rows_event=False,
-                 ignore_decode_errors=False):
+                 ignore_decode_errors=False,
+                 parma_logging=True, ):
```
julien-duponchelle, line 147: I suggest to document it. Maybe we can find a name more explicit also but I have nothing in mind.

julien-duponchelle:
Sorry can you rebase your change agaisnt the last main ?

why-arong:
Before the change, we only considered this parameter as a means to monitor how 'BinLogStreamReader' is configured. However, we felt that there may be additional items to log in this project, so we renamed it to `enable_logging`. For example, the following log entries could be included:
## ERROR
In cases where it fails to pass the CRC check.

## WARN
When the time of event creation differs from the time of table structure retrieval, leading to incorrect mapping.
[Related code](https://github.com/julien-duponchelle/python-mysql-replication/blob/c977dd8dde132acad23511369787bf040f40294b/pymysqlreplication/row_event.py#L655C1-L670C41)

julien-duponchelle:
Good ! I will solve the conflict and merge

julien-duponchelle:
Hmm seem the conflict solve introduce a problem we have lost the changelog of the new version. I'm going to look at that

julien-duponchelle:
Great you fixed it ! So fast

PR #465, opened 2023-09-01
Avoid UnicodeDecodeError for non-utf8 QueryEvents

oseemann:
Query strings in QueryEvents that appear in the binlog stream must not necessarily be utf-8 encoded, but the current implementation handles only utf-8.

This commit adds the `errors="backslashreplace"` kwarg to decode(), to avoid a runtime error and insert \xNN escape sequences for byte sequences that are not valid utf-8. It includes a test that generates a QueryEvent with latin-1 encoding, which fails without the fix.


julien-duponchelle:
That's awesome

PR #468, opened 2023-09-02
Fix GitHub Actions Failure by Updating Test Code for PreviousGtidsEvent

why-arong:
[PreviousGtidsEvent was recently added to the codebase](https://github.com/julien-duponchelle/python-mysql-replication/commit/19fe16c572d184089a13b65e891c59c5f948968d), but the corresponding test cases in the TestBasicBinLogStreamReader class were not updated. This has been causing GitHub Actions to fail.

This PR fixes the issue by updating the test_allowed_event_list method in TestBasicBinLogStreamReader to include the correct expected lengths for _allowed_event_list.

Before the fix, the test expected lengths of 23, 22, and 22 for different scenarios. These have been corrected to 24, 23, and 23, respectively, to align with the addition of PreviousGtidsEvent.

### DIFF
--- a/pymysqlreplication/tests/test_basic.py
+++ b/pymysqlreplication/tests/test_basic.py
@@ -31,9 +31,9 @@ class TestBasicBinLogStreamReader(base.PyMySQLReplicationTestCase):
         return [GtidEvent, PreviousGtidsEvent]
 
     def test_allowed_event_list(self):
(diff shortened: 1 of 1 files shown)


julien-duponchelle:
Oh right I solved conflicts by hand and didn't pay attention thanks !

PR #446, opened 2023-08-25
Mysql 8.0.14 version support Table map Event optional metaData 

sean-k1:
Fixed: #444
This PR depends on : #440 , #477 

Description
---
Mysql 8.0.14 version Table map event optional meta data can be used
when the global variable **binlog_row_metadata** is set to **'FULL'.** AND **binlog_row_image** is set to **FULL**
With this global variable value, we can retrieve column information from the binlog when **a table map event occurs**.

We apologize to those using MYSQL 5, but we no longer support python-mysql-replication prior to version 0.44.0.

If We sync from optional meta data, **Column name Information Flag = True**
스크린샷 2023-09-19 오후 10 29 48

Changes
---
Parsing Column Infromation from  optional metadata

We no longer need to rely on the **current moment's database table schema information**.
We can sync column data at the time of event occurrence

Test
---
I Added Testcase mysql, mariadb version

example Sql
```
CREATE DATABASE test;
use test;
CREATE TABLE test4 (id int NOT NULL AUTO_INCREMENT, data VARCHAR(255), data2 VARCHAR(255), PRIMARY KEY(id));
INSERT INTO test4 (data,data2) VALUES ("Hello", "World");
UPDATE test4 SET data = "World", data2="Hello" WHERE id = 1;
DELETE FROM test4 WHERE id = 1;
```
dump 
```
=== QueryEvent ===
Date: 2023-09-19T13:58:23
Log position: 101256
Event size: 85
Read bytes: 85
Schema: b'test'
Execution time: 0
Query: CREATE DATABASE test

=== QueryEvent ===
Date: 2023-09-19T13:58:23
Log position: 101530
Event size: 172
Read bytes: 172
Schema: b'test'
Execution time: 0
Query: CREATE TABLE test4 (id int NOT NULL AUTO_INCREMENT, data VARCHAR(255), data2 VARCHAR(255), PRIMARY KEY(id))

=== QueryEvent ===
Date: 2023-09-19T13:58:23
Log position: 101684
Event size: 52
Read bytes: 52
Schema: b'test'
Execution time: 0
Query: BEGIN

=== TableMapEvent ===
Date: 2023-09-19T13:58:23
Log position: 101768
Event size: 61
Read bytes: 61
Table id: 233
Schema: test
Table: test4
Columns: 3
=== OptionalMetaData ===
unsigned_column_list: [False, False, False]
default_charset_collation: 255
charset_collation: {}
column_charset: []
column_name_list: ['id', 'data', 'data2']
[...]

PR #471, opened 2023-09-06
Feature/optional meta data extraction

sean-k1:
[The PR](https://github.com/julien-duponchelle/python-mysql-replication/pull/446) that tries to sync column information based on optional meta data was too big, so I separated the part that extracts the optional meta data extraction

Description
---
Mysql 8.0.14 version Table map event optional meta data can be used
when the global variable **binlog_row_metadata** is set to **'FULL'.**
With this global variable value, we can retrieve column information from the binlog when **a table map event occurs**.

Changes
---
obtain table-related information at the time of an event occurrence from optional metadata.

Test
---
I Added Optional meta data extraction Testcase mysql, mariadb version

Optional meta Data dumps like this

스크린샷 2023-09-07 오전 7 59 58


julien-duponchelle:
That's great !

PR #476, opened 2023-09-15
add py-mysql-elasticsearch-sync to readme, a project using pymyrepl

jaehyeonpy:
-  python-mysql-replication has its usage of 'MySQL to search engine replication', but there is only one usage that can be found in examples/logstash.
- The logstash thing is also a good usage, but I found it better to have more usage. So, I made py-mysql-elasticsearch-sync, which replicates mysql to elasticsearch using python-mysql-replication(https://github.com/jaehyeonpy/py-mysql-elasticsearch-sync).
- I would appreciate if you could merge this.


julien-duponchelle:
Great !

PR #477, opened 2023-09-15
Feature/remove column schema

sean-k1:


sean-k1:
@julien-duponchelle 

Hi Julien! 
I'd like to ask you to think deeply about this issue.
By solving this, we can become a more reliable open source project.

julien-duponchelle:
Maybe we could keep the old mode and use the new one if the database is compatible even if it's create a lot of additional complexity. I think it will be usefull for people using the library for moving from old MySQL deployment.

I agree that the new mode is better.

dongwook-chan:
@julien-duponchelle @sean-k1 
Considering the depth of this change and its impact on the project's core functionality, I believe this introduces a new chapter for python-mysql-replication. With that in mind, I'd recommend considering this for a major version increment. This would help signal to the users the significance of the change and set the right expectations regarding compatibility and functionality differences.

julien-duponchelle:
Agree a major version increase is a good idea

julien-duponchelle:
I'm not strongly attached to backward compatibility with MySQL 5 even if it will be ideal. If it's too much effort we can make it clear in the documentation that old version require an old version of the library.

sean-k1:
@julien-duponchelle 
I Check this one 
the database is compatible By SHOW VARIABLES LIKE 'BINLOG_ROW_METADATA';

if this one is None: 
user use Mysql 5 Version Or Before Mariadb 10.5 Version

julien-duponchelle:
Great

Let me know when you are ready for a merge

sean-k1:
@julien-duponchelle I'm ready to merge this PR.
If this PR merged, I resolve the conflict [sync Column Pr](https://github.com/julien-duponchelle/python-mysql-replication/pull/446)

sean-k1:
@julien-duponchelle 
It would be nice if this PR was also merged! 
This PR should have been merged before the column sync PR. 
but as the order of merges changed, the comment didn't change all the way.

julien-duponchelle:
And voilà

PR #478, opened 2023-09-15
Remove duplicated Affected columns output in UpdateRowsEvent

mjs1995:
### Overview

This Pull Request aims to remove the duplicate "Affected columns" output in the `UpdateRowsEvent` class.

### Description

The `_dump()` method in both `RowsEvent` and `UpdateRowsEvent` classes includes a print statement for the number of affected columns. The duplicate output appears when using `UpdateRowsEvent`.

### Changes

- Removed the line `print("Affected columns: %d" % self.number_of_columns)` from the `_dump()` method in the `UpdateRowsEvent` class.

Is there a specific reason for the duplicate "Affected columns" output? I'd appreciate any feedback on this.

### DIFF
--- a/pymysqlreplication/row_event.py
+++ b/pymysqlreplication/row_event.py
@@ -602,7 +602,6 @@ class UpdateRowsEvent(RowsEvent):
 
     def _dump(self):
         super()._dump()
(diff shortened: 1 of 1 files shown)


julien-duponchelle:
No good reason, thanks !

PR #491, opened 2023-09-29
We need a logic to convert events into JSON format.

chungeun-choi:
### Overview

In the majority of projects utilizing the `pymysqlreplication` library, it has been observed that binlog events are being converted into JSON format for usage.

[Examples]

[Streaming Changes in a Database with Amazon Kinesis](https://aws.amazon.com/blogs/database/streaming-changes-in-a-database-with-amazon-kinesis/)

[Near Zero Downtime Migration from MySQL to DynamoDB](https://aws.amazon.com/ko/blogs/big-data/near-zero-downtime-migration-from-mysql-to-dynamodb/)

Consequently, we have developed logic within our library to provide binlog events in JSON format.

- How to execute
    
    ```python
    stream = BinLogStreamReader(
            connection_settings=MYSQL_SETTINGS, server_id=3, blocking=True
        )
    
        for binlogevent in stream:
            json_result = binlogevent.to_json()
            print(json_result)
    ```
    
- Results
    
    ```python
     {
    	"evnet_type": 30,
    	"date": "2023-09-29T04:15:11",
    	"log_pos": 994,
    	"event_size": 103,
    	"read_bytes": 12,
    	"table": ["test", "r1"],
    	"affected_columns": 3,
    	"changed_rows": 7,
    	"flag": true,
    	"event_name": "WriteRowsEvent",
    	"rows": [{
    		"values": {
    			"i1": 1,
    			"c1": "#1",
    			"d1": "2023-09-29T04:15:11"
    		}
    	}, {
    		"values": {
    			"i1": 2,
    			"c1": "#2",
    			"d1": "2023-09-29T04:15:11"
    		}
    	}, {
    		"values": {
    			"i1": 3,
    			"c1": "#3",
    			"d1": "2023-09-29T04:15:11"
    		}
    	}, {
    		"values": {
    			"i1": 4,
    			"c1": "#4",
    			"d1": "2023-09-29T04:15:11"
    		}
    	}, {
    		"values": {
    			"i1": 5,
    			"c1": "#5",
    			"d1": "2023-09-29T04:15:11"
    		}
    	}, {
    		"values": {
    			"i1": 6,
    			"c1": "#6",
    			"d1": "2023-09-29T04:15:11"
    		}
    	}, {
    		"values": {
    			"i1": 7,
    			"c1": "#7",
    			"d1": "2023-09-29T04:15:11"
    		}
    	}]
    }
    ```
    

### Benefit

Through this logic, users can reduce the steps needed for additional processing.

[...]

PR #494, opened 2023-10-02
v0.45 bug fix json object and array

sean-k1:
We're releasing v 0.45 before v 1.0 to fix a json parsing error that mysql 5.7 users are facing.

* bug fix json Parser

* make bytes util

* error fix json array, json object

* add datetime, time parsing add test datetime,decimal,time cases


dongwook-chan:
Seems to be the same issue with your [previous review](
https://github.com/julien-duponchelle/python-mysql-replication/pull/491#discussion_r1341350325)

sean-k1:
@dongwook-chan 
yes Moving the byte util to a separate location seems to be the right solution.

PR #511, opened 2023-10-06
Parse ALTER TABLE ADD/DROP COLUMN statement and sync `table_map`

dongwook-chan:
I've noticed that a recent change in the repository removed the ability to select from information_schema to retrieve column_schema. Instead, it was chosen to parse the optional metadata present in the table map event to obtain this schema. I understand the rationale behind this, but there's a caveat: if binlog_row_metadata is set to FULL, the optional metadata isn't always fully available. This can lead to incomplete or incorrect column schemas.

To address this gap, I've developed a feature that parses the DDL from the query event and updates the table_map accordingly. This ensures that even when binlog_row_metadata is set to FULL, we can accurately and consistently retrieve the column schema.

**Changes Made:**

Implemented DDL parsing from query events.
Updated table_map to reflect changes from parsed DDL.
Added necessary unit tests to ensure correctness and stability of the new feature.
Benefits:

Ensures that column schemas are always accurate, even when binlog_row_metadata is set to FULL.
Reduces reliance on optional metadata which might be incomplete.
I'd appreciate it if you could review the changes and provide feedback. I believe this enhancement can add robustness to the way column schemas are determined, ensuring accuracy and consistency for all configurations.


sean-k1:
@dongwook-chan Could you change this pr to draft?

dongwook-chan:
Sure! 😁😁

PR #512, opened 2023-10-09
refactor : changed formatted string to f-string

mirageoasis:
since python2 has been deprecated in this project I changed string to f-string which increases readablity


dongwook-chan (approved):
Added trailing newline and removed trailing whitespace too!
This is not only more readable but also more modern approach to format string.

PR #515, opened 2023-10-12
Remove python2 and mysql5.7 code

mirageoasis:
while looking around repository I came across python2, mysql 5.7 tests, codes
and Github Actions which was deprecated in python-mysql-replication version 1.0

However, I think there should be some improvement over config.json and docker-compse.yml 

Since config.json port setting depends on docker-compose file.


pymysqlreplication/tests/test_data_type.py
```diff
@@ -287,55 +284,6 @@ def test_timestamp(self):
...
-                event.rows[0]["values"]["test3"],
-                datetime.datetime(1984, 12, 3, 12, 33, 7, 123000),
-            )
-            self.assertEqual(
-                event.rows[0]["values"]["test4"],
-                datetime.datetime(1984, 12, 3, 12, 33, 7, 123400),
-            )
-            self.assertEqual(
-                event.rows[0]["values"]["test5"],
-                datetime.datetime(1984, 12, 3, 12, 33, 7, 123450),
```
sean-k1, line 332: not possible  Test Mysql 8 version?

sean-k1 (commented):
If you delete all mysql 5.7 version Test
it would be better set global variable binlog_row_metadata = "FULL" in setUp method.

mirageoasis:
> If you delete all mysql 5.7 version Test it would be better set global variable binlog_row_metadata = "FULL" in setUp method.

understood I will apply this in following commit

dongwook-chan:
We cannot overlook the MySQL 5.7 user base. I know MySQL declared EOL but there are plenty of users that can benefit from the code.

PR #518, opened 2023-10-13
python2 minor fix

mirageoasis:
In this pr I think change was too radical
[link](https://github.com/julien-duponchelle/python-mysql-replication/pull/515)

so I only remove python2 compatibliity and did some minor fixes


pymysqlreplication/tests/benchmark.py
```diff
@@ -25,12 +25,12 @@ def consume_events():
         only_events=[UpdateRowsEvent],
         only_tables=["test"],
     )
-    start = time.clock()
+    start = time.perf_counter()
```
sean-k1, line 28: perf_counter method is it possible 3.3, 3.4, 3.5 and 3.6?

sean-k1 (commented):
Everything looks good except for this
Can you check?

pymysqlreplication/tests/benchmark.py
mirageoasis: > python

https://docs.python.org/3/library/time.html#time.perf_counter

yes it says this method was introduced in python 3.3

pymysqlreplication/tests/base.py
```diff
@@ -84,6 +84,13 @@ def isMySQL8014AndMore(self):
             return True
         return version == 8.0 and version_detail >= 14
 
+    def isMySQL8016AndMore(self):
+        version = float(self.getMySQLVersion().rsplit(".", 1)[0])
```
dongwook-chan, line 88: Any reasons for opting to cast to float instead of int?

mirageoasis: Because Other code converts version into float I really don't know
Can I investigate this problem?

dongwook-chan: Yes. I believe casting to `int` is enough.
We should apply Boy Scout Rule.
Please investigate.

dongwook-chan: My bad. I mistook the number of split tokens.

PR #547, opened 2023-10-15
add: send mysql_to_rabbitmq.py

### LINKED ISSUE #545: Add Send an Event MySQL to RabbitMQ
Add codes to send to RabbitMQ when a mysql event is triggered

myminju:
closes #545

### DIFF
--- a/examples/mysql_to_rabbitmq.py
+++ b/examples/mysql_to_rabbitmq.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python
+
+#
+# Update a RabbitMQ when an event is triggered
+# in MySQL replication log
+#
+
+import json
+import pika
+
+from pika import DeliveryMode
+from pymysqlreplication import BinLogStreamReader
+from pymysqlreplication.row_event import (
+    DeleteRowsEvent,
+    UpdateRowsEvent,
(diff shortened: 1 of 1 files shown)


examples/mysql_to_rabbitmq.py
```diff
@@ -0,0 +1,67 @@
...
+                routing_key = "order"
+                message_body = row["values"].items()
+
+            elif isinstance(binlogevent, UpdateRowsEvent):
+                routing_key = "order"
+                message_body = row["after_values"].items()
+                
+            elif isinstance(binlogevent, WriteRowsEvent):
+                routing_key = "order"
+                message_body = row["values"].items()
```
sean-k1, line 54: better use routing_key just once.
and I think routing_key == order is little bit Doesn't fit.

dongwook-chan:
I'm gonna test this on local RabbitMQ

dongwook-chan:
I have tested your code thoroughly.
It works fine in general.
However in cases where the column types are temporal (i.e. timestamp or datetime) or JSON,
I got serialization issues.
So I updated to code so that the incompatible column values can be cast to `str` before being serialized to JSON.

Issue #520, opened 2023-10-14
Unknown variable


sean-k1:
Description
---
 row_event.py 89Lines

```
        # etc
        else:
            self.extra_data = self.packet.read(self.extra_info_length - 3)
```
I think It is really rare case.
but we should change extra_info_length => extra_data_length

We don't have this extra_info_length value.
May be typo Issue..

swatichauhan814:
Hi, 
I would like to work on this ? Can you assign this to me ?

sean-k1:
@swatichauhan814 
Sure Thanks!

PR #548, opened 2023-10-15
fix : update unknown variable

swatichauhan814:
closes [520](https://github.com/julien-duponchelle/python-mysql-replication/issues/520)


sean-k1 (approved):
Ignore Workflow bug
Thanks your hard work!🎉

PR #553, opened 2023-10-15
Replace texts with String constants

### LINKED ISSUE #552: Fix TestCode Using Constant instead String
Comparing through constants is better than comparing strings directly

https://github.com/julien-duponchelle/python-mysql-replication/blob/c639f4714956816e3f9b529596e2268caef8daac/pymysqlreplication/tests/test_basic.py#L1793C3-L1793C3

"null" -> NONE_SOURCE.NULL

https://github.com/julien-duponchelle/python-mysql-replication/blob/c639f4714956816e3f9b529596e2268caef8daac/pymysqlreplication/tests/test_basic.py#L1826-L1838

"null" -> NONE_SOURCE.NULL
"out of datetime2 range" -> NONE_SOURCE.OUT_OF_DATETIME2_RANGE
"out of datetime range" -> NONE_SOURCE.OUT_OF_DATETIME_RANGE

sotiriszogos:
Fixes #552

### DIFF
--- a/pymysqlreplication/tests/test_basic.py
+++ b/pymysqlreplication/tests/test_basic.py
@@ -7,10 +7,12 @@ from pymysqlreplication import BinLogStreamReader
 from pymysqlreplication.gtid import GtidSet, Gtid
 from pymysqlreplication.event import *
 from pymysqlreplication.constants.BINLOG import *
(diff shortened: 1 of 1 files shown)


sean-k1 (approved):
Thanks your hardWork! LGTM

Issue #560, opened 2023-10-16
Add Example for MySQL to Kafka events


HongGeonUi:
### Issue Description

Add Example for MySQL to Kafka events.

### Expected Outcome

We can use MySQL to Kafka events with example.

PR #561, opened 2023-10-16
Add : Example for MySQL to Kafka events

HongGeonUi:
In #560
Add Example for MySQL to Kafka Events.

It's only provide DeleteRowEvent, UpdateRowEvent, WriteRowEvent.
This is the result of example.

스크린샷 2023-10-17 오전 12 33 33

### DIFF
--- a/examples/mysql_to_kafka.py
+++ b/examples/mysql_to_kafka.py
@@ -1,0 +1,86 @@
+#!/usr/bin/env python
+
+#
+# Output Kafka events to the console from MySQL replication stream
+#
+
+from pymysqlreplication import BinLogStreamReader
+from pymysqlreplication.row_event import (
+    DeleteRowsEvent,
+    UpdateRowsEvent,
+    WriteRowsEvent,
+)
+
+from kafka.admin import KafkaAdminClient, NewTopic
+from kafka import KafkaConsumer, KafkaProducer
+
+MYSQL_SETTINGS = {"host": "127.0.0.1", "port": 3306, "user": "root", "passwd": ""}
+
+
+def create_kafka_topic(topic_name, num_partitions=1, replication_factor=1):
+    admin_client = KafkaAdminClient(bootstrap_servers="127.0.0.1:9092")
+
+    topic = NewTopic(
+        name=topic_name,
+        num_partitions=num_partitions,
+        replication_factor=replication_factor
+    )
+
+    admin_client.create_topics(new_topics=[topic])
+
+
+def main():
+    global message_body
+    producer = KafkaProducer(
+        bootstrap_servers='127.0.0.1:9092',
+        value_serializer=lambda v: str(v).encode('utf-8')
+    )
+
+    stream = BinLogStreamReader(
+        connection_settings=MYSQL_SETTINGS,
+        server_id=3,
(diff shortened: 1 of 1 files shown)


dongwook-chan:
Well done!
I'm impressed that you pulled this off despite insufficient instructions on the project.
I appreciate your effort to improve the project.
Let me just run some tests before I merge.

examples/mysql_to_kafka.py
```diff
@@ -0,0 +1,84 @@
...
+
+            elif isinstance(binlogevent, UpdateRowsEvent):
+                topic = "updated"
+                message_body = row["after_values"].items()
+
+            elif isinstance(binlogevent, WriteRowsEvent):
+                topic = "created"
+                message_body = row["values"].items()
+
+            producer.send(topic, key=None, value=message_body)
```
sean-k1, line 60:
```suggestion
            producer.send(topic, key=None, value=dict(message_body))
```
I think wrapping dict or json format is more better.

dongwook-chan: Yes. Current code might also impose some serialization issue. I will run some test thoroughly to prevent potential issues.

HongGeonUi: I also think it's good for serialization. I wrap that message_body
I have one question.
In current code, I close stream, producer, consumer.
So I can only get the message before the main() execution.
Is that any problem?

dongwook-chan: I think you did a good job on that one.
My expectations were that you would implementing only producing of binlogevents.
Yet, you managed to implement the whole package. 🙌🏻🙌🏻🙌🏻
I think your code is straightforward and it wouldn't be any problem at all.

I'll get back to you with the serialization issue after my test is done.

dongwook-chan (changes requested):

examples/mysql_to_kafka.py
```diff
@@ -0,0 +1,83 @@
...
+        "deleted",
+        "updated",
+        "created",
+        bootstrap_servers="127.0.0.1:9092",
+        value_deserializer=lambda x: x.decode("utf-8"),
+        auto_offset_reset="earliest",
+        group_id="1",
+    )
+
+    for message in consumer:
```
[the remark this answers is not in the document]
HongGeonUi: Ok, I handle with KeyboardInterrupt.
Additional, I found one more error in process of closing like this.

스크린샷 2023-10-18 오후 1 47 28

It is issue in Kafka-python https://github.com/dpkp/kafka-python/issues/1583
So I add time.sleep(1) under producer.close().
I think it's not the best way to handle this error. but I don't know how to handle this with other way.
Isn't this a problem?

Finally, I think we handle all the errors occur when we use this example.
Thx for your feedback and support.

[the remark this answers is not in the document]
dongwook-chan: The purpose of example codes are to showcase the use case.
In this case, it shows how you can incorporate kafka-python.
Thanks to you, this is already fulfilled.
So any other details that are specific to the external dependency can be left out.
Going top deep with the exception may hinder newcomers' understanding of the examples.
Catching BaseException or KeyboardInterrupt IS NOT a perfect way to detecting end of the program
but it can hint the users that the code ends at that point and I think that's enough for the example code.
So great job!

dongwook-chan (approved):

PR #563, opened 2023-10-18
Add error logging for crc32 validation failures.

why-arong:
### Description

This PR add logging for situations where CRC32 validation does not pass.

By monitoring these failures, we can ensure data integrity throughout the application.

### DIFF
--- a/pymysqlreplication/event.py
+++ b/pymysqlreplication/event.py
@@ -58,10 +58,8 @@ class BinLogEvent(object):
         data = self.packet.read(19 + self.event_size)
         footer = self.packet.read(4)
         byte_data = zlib.crc32(data).to_bytes(4, byteorder="little")
-        if byte_data == footer:
-            self._is_event_valid = True
-        else:
-            self._is_event_valid = False
+        self._is_event_valid = True if byte_data == footer else False
+        if not self._is_event_valid:
             logging.error(
                 f"An CRC32 has failed for the event type {self.event_type}, "
                 "indicating a potential integrity issue with the data."
--- a/pymysqlreplication/event.py
+++ b/pymysqlreplication/event.py
@@ -3,6 +3,7 @@ import struct
 import datetime
 import decimal
 import zlib
+import logging
 
 from pymysqlreplication.constants.STATUS_VAR_KEY import *
 from pymysqlreplication.exceptions import StatusVariableMismatch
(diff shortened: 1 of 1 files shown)


pymysqlreplication/event.py
```diff
@@ -57,7 +58,14 @@ def _verify_event(self):
...
         byte_data = zlib.crc32(data).to_bytes(4, byteorder="little")
-        self._is_event_valid = True if byte_data == footer else False
+        if byte_data == footer:
+            self._is_event_valid = True
+        else:
+            self._is_event_valid = False
+            logging.error(
+                f"An CRC32 has failed for the event type {self.event_type}, "
+                "indicating a potential integrity issue with the data."
+            )
```
sean-k1, line 68:
```suggestion
      self._is_event_valid = True if byte_data == footer else False
      if not self._is_event_valid:
            logging.error(
                f"An CRC32 has failed for the event type {self.event_type}, "
                "indicating a potential integrity issue with the data."
            )
```
@why-arong 
I think this looks more readable, what do you think?
I think the code you wrote is good, too.

why-arong: I think that code reads better!

How about error message? Is there any additional information we can provide besides `'self.event_type'`??

why-arong:
https://github.com/julien-duponchelle/python-mysql-replication/blob/1d2c8aba5d96ca4573054e51038d9f62b43da9e9/pymysqlreplication/binlogstream.py#L553-L558

How do you feel about not writing the code like `logging.log(logging.ERROR, ~~~)`? Unless I need to set the log level dynamically, I think `logging.error(~~~)` is more intuitive, so that's how I wrote it!

pymysqlreplication/event.py
sean-k1: I think that's enough

sean-k1:
> @sean-k1 
> 
> https://github.com/julien-duponchelle/python-mysql-replication/blob/1d2c8aba5d96ca4573054e51038d9f62b43da9e9/pymysqlreplication/binlogstream.py#L553-L558
> 
> 
> 
> How do you feel about not writing the code like `logging.log(logging.ERROR, ~~~)`? Unless I need to set the log level dynamically, I think `logging.error(~~~)` is more intuitive, so that's how I wrote it!

logging.error is more better👍

sean-k1 (approved):
LGTM Thanks your hard work!

PR #569, opened 2023-10-20
refactor : eradicate f-string

mirageoasis:
I tracked down formatted string and changed ramaining strings

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -81,11 +81,11 @@ class ReportSlave(object):
             self.hostname = value
 
     def __repr__(self):
-        return "<ReportSlave hostname=%s username=%s password=%s port=%d>" % (
-            self.hostname,
-            self.username,
-            self.password,
-            self.port,
+        return (
+            f"<ReportSlave hostname={self.hostname} "
+            f"username={self.username} "
+            f"password={self.password} "
+            f"port={self.port}>"
         )
 
     def encoded(self, server_id, master_id=0):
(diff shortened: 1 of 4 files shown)


mirageoasis:
When I used f-string it cannot call pymysql.cursor._escape_args which cause error in process

pymysqlreplication/binlogstream.py
```diff
@@ -367,7 +367,7 @@ def __connect_to_stream(self):
             # master_heartbeat_period is nanoseconds
             heartbeat = int(heartbeat * 1000000000)
             cur = self._stream_connection.cursor()
-            cur.execute("SET @master_heartbeat_period= %d" % heartbeat)
+            cur.execute(f"SET @master_heartbeat_period= {heartbeat}")
```
dongwook-chan, line 370: This is not the recommended way to use DB API in Python. First argument of `execute` must be a query string with placeholders %s for the following arguments to be replaced with.

https://peps.python.org/pep-0249/#id20

dongwook-chan (changes requested):

pymysqlreplication/binlogstream.py
mirageoasis: > This is not the recommended way to use DB API in Python. First argument of `execute` must be a query string with placeholders %s for the following arguments to be replaced with.
> 
> https://peps.python.org/pep-0249/#id20

thanks for a good point! if were not you I could not have recognized this recommandation

julien-duponchelle:
I like this upgrade to modern Python !

PR #570, opened 2023-10-20
Fix: Apply f-string binlogstream.py L135:L141

HongGeonUi:
I found string format in binlogstream.py.
So I convert to f-string that lines.


dongwook-chan (approved):
Great catch! Thanks for submitting the PR. You've made the code more readable and modern. Please don't hesitate to create a PR and ask any questions.

PR #579, opened 2023-11-23
fake rotate event check

### LINKED ISSUE #578: A wrong warning message is logged for every event from ignored table
### Version
Please specify the versions you are using. Exact version numbers are preferred.
- Pymyrepl 1.0.3
- OS Ubuntu 22.04
- Database and version (Remove unnecessary options):
  - MySQL 8.0.35

### System Variables
List relevant system variables using the query `SHOW VARIABLES LIKE '<variable_name_here>';`
- `log_bin`: on
- `binlog_format`: row
- `binlog_row_image`: full
- `binlog_row_metadata`: full
- `enforce_gtid_consistency`: on
- `gtid_mode`: on

If you are using MySQL 8.0.14 or later, please also provide the following variables:
- `binlog_row_metadata`: on
- `binlog_row_value_options`: default

### Symptoms
The message is `A pymysql.OperationalError error occurred, causing a fake rotate event and initialization of the table_map`.
But there is no fake rotate event.

### Steps to Reproduce
Initialize `BinLogStreamReader` with `only_tables=[#some tables,...]` and start iterating it.

### Expected Outcome
No warning is logged.

### Actual Outcome
Then for each event from a table that is not in the table list, the warning is logged.

### My interpretation of the issue
The issue is caused by this PR: https://github.com/julien-duponchelle/python-mysql-replication/pull/490
If I understand correctly, the `table_map` should be empty when we have a fake rotate event.
But here, it is not empty. It just doesn't have a key for the table.
It is possible, too, that the table is empty, but that's because no events have been received yet for the tables we are interested in.
I think a flag for the rotation is needed.

### Planning to Resolve

- [ ] I plan to create a pull request to resolve this issue.
- [x] I don't plan to resolve this myself and would like someone else to address it.

sean-k1:
### Description
resolve #578 

We initialize the table_map when we create the binlogstreamreader object.

After that, if the binlog is maxsize full, a rotateEvent is fired.
At that point, we empty the table_map because the timestamp is not zero.
Then, when we connect to the next binlogfile, we get a rotateEvent with a timestamp of 0. We don't need to empty the table_map again.

However, when we disconnect from mysql and reconnect, we get a fake rotateEvent with a 0 timestamp.
In this case, we inform the client of error logging and do not initialize the table map.

### Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

### Checklist
- [ ] This PR resolves an issue #[Issue Number Here].

### Tests
- [x] All tests have passed.

### DIFF
--- a/pymysqlreplication/binlogstream.py
+++ b/pymysqlreplication/binlogstream.py
@@ -596,6 +596,11 @@ class BinLogStreamReader(object):
                 if code in MYSQL_EXPECTED_ERROR_CODES:
                     self._stream_connection.close()
                     self.__connected_stream = False
+                    logging.WARN(
+                        """
+                          A pymysql.OperationalError error occurred, Re-request the connection.
+                        """
+                    )
                     continue
                 raise
 
@@ -635,7 +640,9 @@ class BinLogStreamReader(object):
                 # invalidates all our cached table id to schema mappings. This means we have to load them all
                 # again for each logfile which is potentially wasted effort but we can't really do much better
                 # without being broken in restart case
-                self.table_map = {}
+                if binlog_event.timestamp != 0:
+                    self.table_map = {}
+
             elif binlog_event.log_pos:
                 self.log_pos = binlog_event.log_pos
 
--- a/pymysqlreplication/row_event.py
+++ b/pymysqlreplication/row_event.py
@@ -1,4 +1,3 @@
-import logging
 import struct
 import decimal
 import datetime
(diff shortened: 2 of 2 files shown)


dongwook-chan (approved):
LGTM

Zaynullin:
Hi there! Can you please repead the changes for the version v0.45.+? I use mysql version 5.7 and caught the same problem. I found that your PR already resolved this problem, but for version 1.0.+.

Issue #334, opened 2020-12-12
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte`


imysm:
MySQL:
     mysql  Ver 8.0.21 for Linux on x86_64 (MySQL Community Server - GPL)
Python:
     Python 3.8.5
mysql-replication:
      mysql-replication                  0.22

Error:
`    for row in binlogevent.rows:
  File "/usr/local/anaconda3/lib/python3.8/site-packages/pymysqlreplication/row_event.py", line 433, in rows
    self._fetch_rows()
  File "/usr/local/anaconda3/lib/python3.8/site-packages/pymysqlreplication/row_event.py", line 428, in _fetch_rows
    self.__rows.append(self._fetch_one_row())
  File "/usr/local/anaconda3/lib/python3.8/site-packages/pymysqlreplication/row_event.py", line 517, in _fetch_one_row
    row["before_values"] = self._read_column_data(self.columns_present_bitmap)
  File "/usr/local/anaconda3/lib/python3.8/site-packages/pymysqlreplication/row_event.py", line 132, in _read_column_data
    values[name] = self.__read_string(1, column)
  File "/usr/local/anaconda3/lib/python3.8/site-packages/pymysqlreplication/row_event.py", line 224, in __read_string
    string = string.decode(encoding)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte`

jhelbig:
This is probably due to invalid characters existing in your database and your character encoding set on your database/table isn't able to process it.

The logic itself is actually pretty awesome, extracting the output encoding to meet the needs of the DB tables, but it's very frustrating because there's no way to manually set the decode format(at least none that I can see).

This is the line that's causing the(or similar) problems.

https://github.com/noplay/python-mysql-replication/blob/main/pymysqlreplication/row_event.py#L249

If there was a way to set a manual encoding value and default to the DB provided value I think it might be a more flexible solution.

tomasluken:
I was having the same error and was able to work around it by changing the default encoding of the mysql database. ( --character-set-server=utf8 --collation-server=utf8_unicode_ci). 
I read that the default changed in version 8 so thats worth checking.

YAtOff:
The issue happens for me with `BLOB` data type columns. Blobs shouldn't be decoded. They store arbitrary binary data. Removing the decoding for blobs fixes the issue.
```patch
[...]

PR #582, opened 2023-11-29
Fix an issue in parsing default column charset

YAtOff:
### Description
  When parsing the default charset optional metadata, when handling the column index/collation number pairs, the column index is in the range of the number of columns with charset (not the number of all the columns).
  Like how it's done here: https://github.com/ClickHouse/ClickHouse/blob/001b67863a57c4d734d94e375e6fdeb4a9407951/src/Core/MySQL/MySQLReplication.cpp#L353
  
  ### Type of Change
  - [x] Bug fix
  
  ### Checklist
  - [x] This PR resolves an issue #[334](https://github.com/julien-duponchelle/python-mysql-replication/issues/334).
  
  ### Tests
  - [x] All tests have passed.

### DIFF
--- a/pymysqlreplication/row_event.py
+++ b/pymysqlreplication/row_event.py
@@ -1005,14 +1005,17 @@ class TableMapEvent(BinLogEvent):
         column_type_detect_function,
     ):
         column_charset = []
+        position = 0
         for i in range(self.column_count):
             column_type = self.columns[i].type
             if not column_type_detect_function(column_type, dbms=self.dbms):
                 continue
-            elif i not in column_charset_collation.keys():
-                column_charset.append(default_charset_collation)
             else:
-                column_charset.append(column_charset_collation[i])
+                if position not in column_charset_collation.keys():
+                    column_charset.append(default_charset_collation)
+                else:
+                    column_charset.append(column_charset_collation[position])
+                position += 1
 
         return column_charset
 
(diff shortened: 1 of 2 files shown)


sean-k1:
@YAtOff Can you describe your mysql variable?

- `binlog_format`:
- `binlog_row_image`:
- `binlog_row_metadata`:
- `binlog_row_value_options`:

YAtOff:
MySQL variables

Name|Value
-|-
binlog_format|ROW
binlog_row_image|FULL
binlog_row_metadata|MINIMAL
binlog_row_value_options|-

pymysqlreplication/row_event.py
```diff
@@ -1005,14 +1005,17 @@ def _parsed_column_charset_by_default_charset(
         column_type_detect_function,
     ):
         column_charset = []
+        position = 0
```
sean-k1, line 1008: @YAtOff confused..
if binlog_row_metadata = "MINIMAL" 
we can not enter this method.
because your  self.__optional_meta_data value is False

``
 def _sync_column_info(self):
        if not self.__optional_meta_data: 
            # If optional_meta_data is False Do not sync Event Time Column Schemas
            return
        if len(self.optional_metadata.column_name_list) == 0:
            # May Be Now BINLOG_ROW_METADATA = FULL But Before Action BINLOG_ROW_METADATA Mode = MINIMAL
            return
```

YAtOff: Sorry, my mistake. It's `full`.
But if it's `minimal` the tests pass without asserts.

sean-k1 (approved):

PR #586, opened 2023-12-01
Add properties in gtid event

Bue-von-hon:
### Description

Motivation:
GtidEvent is aleady implemented but there are missing 6 properties.
If we respect these properties, user can use GtidEvent with more convenient.

Motification:
- Fix(event.py): Add two properties.
- Add(test_basic.py): Add test for GtidEvent.

Result:
for now on, user can use read_immediate_commit_timestamp and read_original_commit_timestamp properties.

### Type of Change
- [ ] Bug fix
- [X] New feature
- [ ] Documentation update

### Checklist
- [X] I have read and understood the [CONTRIBUTING.md](https://github.com/julien-duponchelle/python-mysql-replication/blob/main/CONTRIBUTING.md) document.
- [X] I have checked there aren't any other open [Pull Requests](https://github.com/julien-duponchelle/python-mysql-replication/pulls) for the desired changes.
- [X] This PR resolves an issue #[Issue Number Here].

### Tests
- [X] All tests have passed.
- [X] New tests have been added to cover the changes. (describe below if applicable).


sean-k1:
@Bue-von-hon 
Thanks for contributing!
Can you give me the url you referenced to check the code?

Bue-von-hon:
@sean-k1 here is the official mysql documentation and the url of the cpp implementation that I referenced.

Please see the Binary Format section in Mysql: https://dev.mysql.com/doc/dev/mysql-server/8.1.0/classbinary__log_1_1Gtid__event.html

You can see the binary format starting at line 428 of the CPP implementation.
CPP: https://github.com/mysql/mysql-server/blob/ca51ab1529a9ec4d73b368df15b65534485a617d/libbinlogevents/src/control_events.cpp#L417

pymysqlreplication/tests/test_basic.py
```diff
@@ -1473,6 +1472,39 @@ def test_rows_query_log_event(self):
...
+            "CREATE TABLE IF NOT EXISTS test (id INT AUTO_INCREMENT PRIMARY KEY, name VARCHAR(255))"
+        )
+        gtid_event = self.stream.fetchone()
+
+        self.assertIsInstance(gtid_event, GtidEvent)
+        self.assertIsInstance(gtid_event.event_type, int)
+        self.assertIsInstance(gtid_event.sid, bytes)
+        self.assertIsInstance(gtid_event.gno, int)
+        self.assertIsInstance(gtid_event.lt_type, int)
+        self.assertIsInstance(gtid_event.last_committed, int)
```
sean-k1, line 1501: Upper mysql 5.7 version has this variable so pytest failed.
1.  base.py make method **isMySQL57AndMore**
2.  setUp method  use *isMySQL57AndMore** before start Test

Bue-von-hon: Resolves it!

Bue-von-hon:
@sean-k1 Leave a comment about the test. PTAL🙏

Summary:
The test is failing due to a mismatch between the mysql version found in the isMySQL57AndMore method and the mysql version in the __init__ method of the BinLogEvent class.

Details:
Upon investigating the reason for the test failure, I found that the __init__ method of the BinLogEvent class written in event.py is not getting the correct mysql version information.
The __init__ method simply returns (0, 0, 0), which is set to the default value, as the version information.

In the isMySQL57AndMore method of the PyMySQLReplicationTestCase class written in base.py, I am getting the correct mysql version information (8.1 in my case).

So, the reason for the test failure is as follows
The isMySQL57AndMore method passes and the test runs, but because the __init__ method returns mysql incorrect version information,
The following logic written in line 107 of the __init__ method of the GtidEvent class does not set the last_committed, sequence_number properties.

```python
if self.mysql_version >= (5, 7):
    self.last_committed = struct.unpack("<Q", self.packet.read(8))[0]
    self.sequence_number = struct.unpack("<Q", self.packet.read(8))[0]
```

pymysqlreplication/tests/test_basic.py
```diff
@@ -1473,6 +1472,41 @@ def test_rows_query_log_event(self):
...
+        self.execute("SET SESSION binlog_rows_query_log_events=0")
+        super(TestGtidEvent, self).tearDown()
+
+    def test_gtid_event(self):
+        self.stream.close()
+        self.stream = BinLogStreamReader(
+            self.database,
+            server_id=1024,
+            only_events=[GtidEvent],
+        )
```
sean-k1, line 1490: @Bue-von-hon 
may be we did not pass FORMAT_DESCRIPTION_EVENT
so mysql_version is not initalized.
Could you change Like this?
```suggestion
        self.stream = BinLogStreamReader(
            self.database,
            server_id=1024,
            only_events=[GtidEvent, FORMAT_DESCRIPTION_EVENT],
        )
```

pymysqlreplication/tests/test_basic.py
```diff
@@ -1486,7 +1486,7 @@ def test_gtid_event(self):
         self.stream = BinLogStreamReader(
             self.database,
             server_id=1024,
-            only_events=[GtidEvent],
+            only_events=[GtidEvent, FORMAT_DESCRIPTION_EVENT],
```
sean-k1, line 1489:
```suggestion
            only_events=[GtidEvent, FormatDescriptionEvent],
```
sorry we should change like this.

Bue-von-hon: @sean-k1 Applied!

pymysqlreplication/tests/test_basic.py
```diff
@@ -1473,6 +1472,41 @@ def test_rows_query_log_event(self):
...
+            self.database,
+            server_id=1024,
+            only_events=[GtidEvent, FormatDescriptionEvent],
+        )
+        if not self.isMySQL57AndMore():
+            self.skipTest("Mysql version is under 5.7")
+        self.execute(
+            "CREATE TABLE IF NOT EXISTS test (id INT AUTO_INCREMENT PRIMARY KEY, name VARCHAR(255))"
+        )
+        gtid_event = self.stream.fetchone()
```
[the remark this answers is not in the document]
Bue-von-hon: After i grep FormatDescriptionEvent, test was passed in my local env. 👍

sean-k1:
@Bue-von-hon 
https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-1.html
Immediate commit timestamp original commit timestamp are introduced in MySQL-8.0.1
so Test Failed in 5.7 env

Bue-von-hon:
> @Bue-von-hon https://dev.mysql.com/doc/relnotes/mysql/8.0/en/news-8-0-1.html#:~:text=databases.%20(WL%20%234618)-,Replication,-%3A%20An%20infrastructure Immediate commit timestamp original commit timestamp are introduced in MySQL-8.0.1 so Test Failed in 5.7 env

I have read the mysql release notes you attached.
Thanks to @sean-k1, I was able to improve the tests to respect the mysql version information.
I ran the test on my local environment against mysql 8.1 and 5.7 versions and found that it passes.
PTAL 🙏

sean-k1:
@Bue-von-hon 

5.7 users will get an error and terminate the process if they use this version 
I'm asking 5.7 users to write below 1.00, but I don't think it's a good idea to at least have the process terminate in error.

I'm not sure if branching based on myql version is a good idea either, so I'll think about it a bit more. 
@dongwook-chan What do you think?

dongwook-chan:
@Bue-von-hon @sean-k1 
It's always almost best to stick to the implementations in mysql-server.
In that sense, python-mysql-replication equivalent of mysql-server 's [can_read](https://github.com/mysql/mysql-server/blob/dee24346f205e1907e2086e8a8684e77b6e292fe/libs/mysql/binlog/event/control_events.cpp#L521).
We need to check how many bytes we have left to read before common header starts and branch accordingly.

I have already created a similar method [bytes_to_read](https://github.com/julien-duponchelle/python-mysql-replication/blob/6f77f70eaf7225be5bd40ddfbcba4281f09e2d00/pymysqlreplication/packet.py#L385) but I would have to say the implementation isn't necessarily elegant.
So you're welcome to modify it or create new one.

PR #592, opened 2023-12-06
test dump method in test_delete_row_event

roy0424:
### Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Other (add test dump method)

### Checklist
- [ ] This PR resolves an issue #[Issue Number Here].

### Tests
- [x] All tests have passed.


sean-k1:
@roy0424 
The user var event is the user entering the value directly, so I think the mock test makes sense, but the
I don't understand why the delete event generates a separate mock event.
Is there any particular reason?

roy0424:
@sean-k1 I made it to improve coverage. But I realized that this test was unnecessary

PR #598, opened 2024-01-01
Replace the mailling list by github discussions

julien-duponchelle:
The mailing list has a low traffic I propose to replace it by Github discussions because it's more accessible to new people.


dongwook-chan:
Sounds great to me!

dongwook-chan (approved):

Issue #601, opened 2024-01-23
Decode Exception due to Inaccurate Length from self.query_length Leading to Byte Sequence Truncation


zj5220924:
### Version
Please specify the versions you are using. Exact version numbers are preferred.
- Pymyrepl (1.1.0):
- OS (Centos 7):
- Pymyrepl (1.0.6):
- Database and version (Remove unnecessary options):
  - MySQL:  8.0.32

### System Variables
List relevant system variables using the query `SHOW VARIABLES LIKE '<variable_name_here>';`
- `log_bin`: YES
- `binlog_format`: ROW
- `binlog_row_image`: FULL
- `enforce_gtid_consistency`: True
- `gtid_mode`: YES

If you are using MySQL 8.0.14 or later, please also provide the following variables:
- `binlog_row_metadata`: FULL
- `binlog_row_value_options`: Null

error massage:
Traceback (most recent call last):
  File "/ROOT/tmp/my2sql/my2sql.py", line 144, in <module>
    binlog2sql.process_binlog()
  File "/ROOT/tmp/my2sql/my2sql.py", line 72, in process_binlog
    for binlog_event in stream:
  File "/usr/local/python3.11/lib/python3.11/site-packages/pymysqlreplication/binlogstream.py", line 617, in fetchone
    binlog_event = BinLogPacketWrapper(
                   ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/python3.11/lib/python3.11/site-packages/pymysqlreplication/packet.py", line 115, in __init__
    self.event = event_class(
                 ^^^^^^^^^^^^
  File "/usr/local/python3.11/lib/python3.11/site-packages/pymysqlreplication/event.py", line 885, in __init__
    self.query = self.packet.read(self.query_length).decode("utf-8")
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
UnicodeDecodeError: 'utf-8' codec can't decode bytes in position 205-206: unexpected end of data

cat /....../pymysqlreplication/event.py
......
    def __init__(self, from_packet, event_size, table_map, ctl_connection, **kwargs):
        super(RowsQueryLogEvent, self).__init__(
            from_packet, event_size, table_map, ctl_connection, **kwargs
        )
        **self.query_length = self.packet.read_uint8()
        self.query = self.packet.read(self.query_length).decode("utf-8")**
        print(self.query_length)
        print(self.query)
[...]

PR #602, opened 2024-01-23
Fix parsing of query field of RowsQueryEvent

dongwook-chan:
### Description
The first byte of payload in RowsQueryEvent must be ignored without being used as the length for the following field `query`. Yet, the existing code did use this value as length for `query` and  fails whenever parsing query longer than 255 characters. 

I fixed the bug referring to 
[code from msql-server](https://github.com/mysql/mysql-server/blob/dee24346f205e1907e2086e8a8684e77b6e292fe/libs/mysql/binlog/event/rows_event.cpp#L537-L543).

Other clients in different languages implement the RowsQueryEvent the way mysql-server does and noted below as references.

[MysqlCdc (c#)](https://github.com/HongGeonUi/MySqlCdc/blob/45328aa819d8975054d632217cad5c375b1d13d1/src/MySqlCdc/Providers/MySql/Parsers/RowsQueryEventParser.cs#L18)
[mysql-binlog-connector-java (java)](https://github.com/osheroff/mysql-binlog-connector-java/blob/f6b853ba62faad5ff59d6aa09856761878484eb1/src/main/java/com/github/shyiko/mysql/binlog/event/deserialization/RowsQueryEventDataDeserializer.java#L31)

### Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Documentation update

### Checklist
- [x] This PR resolves an issue #601.

### Tests
- [x] All tests have passed.

### DIFF
--- a/pymysqlreplication/event.py
+++ b/pymysqlreplication/event.py
@@ -880,12 +880,11 @@ class RowsQueryLogEvent(BinLogEvent):
         super(RowsQueryLogEvent, self).__init__(
             from_packet, event_size, table_map, ctl_connection, **kwargs
         )
-        self.query_length = self.packet.read_uint8()
-        self.query = self.packet.read(self.query_length).decode("utf-8")
+        self.packet.advance(1)
+        self.query = self.packet.read_available().decode("utf-8")
 
     def dump(self):
         print(f"=== {self.__class__.__name__} ===")
-        print(f"Query length: {self.query_length}")
         print(f"Query: {self.query}")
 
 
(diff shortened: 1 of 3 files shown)


pymysqlreplication/event.py
```diff
@@ -880,12 +880,11 @@ def __init__(self, from_packet, event_size, table_map, ctl_connection, **kwargs)
         super(RowsQueryLogEvent, self).__init__(
             from_packet, event_size, table_map, ctl_connection, **kwargs
         )
-        self.query_length = self.packet.read_uint8()
-        self.query = self.packet.read(self.query_length).decode("utf-8")
+        self.packet.advance(1)
```
sean-k1, line 883: How about delete line#875

sean-k1 (commented):
except reivew LGTM

pymysqlreplication/event.py
dongwook-chan: My bad. I'll fix this right away. Thx for the thorough review!😁

sean-k1 (approved):
LGTM 
except fix lint

avaitla:
This also crashes for our MySQL 8.035. Would love to see it merged in.

PR #634, opened 2025-11-06
Add MySQL 5.7 column name support

kartalbas:
## Summary

This PR adds support for getting column names when using MySQL 5.7, which doesn't include column names in the binary log by default.

## What Changed

- Added new option `use_column_name_cache` to enable column name fetching
- Fetches column names from `INFORMATION_SCHEMA.COLUMNS` when binlog metadata is missing
- Uses in-memory caching to avoid repeated database queries
- Only runs when explicitly enabled (opt-in feature)

## Why This Is Needed

- MySQL 5.7 does not support `binlog_row_metadata=FULL` setting
- Without this feature, column names are unavailable in binlog events
- Many legacy applications still run on MySQL 5.7 and cannot upgrade due to operational constraints, infrastructure dependencies, or business requirements
- This feature enables these applications to continue using mysql-replication without requiring a database upgrade
- Applications need column names to properly process row change events

## How It Works

- When `use_column_name_cache=True` is set, the library queries `INFORMATION_SCHEMA` for column names
- Results are cached to improve performance
- Falls back gracefully if the query fails
- Default behavior remains unchanged (feature is disabled by default)

## Testing

Tested with MySQL 5.7 databases where binlog metadata is not available.

### DIFF
--- a/README.md
+++ b/README.md
@@ -366,6 +366,7 @@ Other contributors:
 - Oliver Seemann: Handle large json, github actions,
   Zero-pad fixed-length binary fields (https://github.com/oseemann)
 - Mahadir Ahmad: Handle null json payload (https://github.com/mahadirz)
+- Mehmet Kartalbas: Add MySQL 5.7 column name support (https://github.com/kartalbas)
 - Axel Viala: Removal of Python 2.7 (https://github.com/darnuria)
 - Etern: Add XAPrepareEvent, parse last_committed & sequence_number of GtidEvent (https://github.com/etern)
 - Jason Fulghum: typo in ident variable name (https://github.com/fulghum)
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -26,6 +26,7 @@ Contents
 
     installation
     limitations
+    mysql57_support
     binlogstream
     events
     examples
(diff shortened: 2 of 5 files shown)


julien-duponchelle:
Sorry you have a failure in the test :
pkt = <pymysql.protocol.MysqlPacket object at 0x00007f9b9fdcd2f0>

    def create_binlog_packet_wrapper(pkt):
        return BinLogPacketWrapper(
            pkt,
            self.stream.table_map,
            self.stream._ctl_connection,
            self.stream.mysql_version,
            self.stream._BinLogStreamReader__use_checksum,
            self.stream._BinLogStreamReader__allowed_events_in_packet,
            self.stream._BinLogStreamReader__only_tables,
            self.stream._BinLogStreamReader__ignored_tables,
            self.stream._BinLogStreamReader__only_schemas,
            self.stream._BinLogStreamReader__ignored_schemas,
            self.stream._BinLogStreamReader__freeze_schema,
            self.stream._BinLogStreamReader__ignore_decode_errors,
            self.stream._BinLogStreamReader__verify_checksum,
>           self.stream._BinLogStreamReader__optional_meta_data,
        )
E       TypeError: __init__() missing 1 required positional argument: 'enable_logging'

test_basic.py:621: TypeError

(I know  it's not easy to run the test :(  )

julien-duponchelle:
Thanks a lot !

